PR #9, opened 2020-07-09
Add route to operator

slagle:
Add the creation of the Keystone route to the operator reconcile method
instead of using the manual yaml file.


pkg/keystone/route.go
```diff
@@ -0,0 +1,35 @@
...
+
+	labels := map[string]string{
+		"app": "keystone-api",
+	}
+	serviceRef := routev1.RouteTargetReference{
+		Kind: "Service",
+		Name: "keystone",
+	}
+	routePort := &routev1.RoutePort{
+		TargetPort: intstr.FromInt(5000),
```
dprince, line 21: I had done this slightly differently. Instead of hard coding the port here I referenced the: targetPort: api in the service (which already has port 5000 hard coded). Either way is probably fine I guess as I'd grep for it.

PR #14, opened 2020-07-16
Add Keystone endpoint and service handling

slagle:


mdbooth:
The KeystoneEndpoint stuff I had down as 'nice to have, probably never do it', but it looks clean and simple. I like it.

pkg/controller/keystoneservice/keystoneservice_controller.go
```diff
@@ -0,0 +1,163 @@
...
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileKeystoneService) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling KeystoneService")
+
+	// Fetch the KeystoneService instance
+	instance := &keystonev1.KeystoneService{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
```
dprince, line 91: I think we'll also want to put something in here to block and Requeue until Keystone bootstrap is completed. Otherwise requests could just fail. During initial deployment we could have multiple services trying to create endpoints, etc at once so blocking until that completes would be ideal. You could look for the BootStrapHash which we are recording in the keystone status objects. If that is set I think it would be safe to continue here as that means the bootstrap job has completed.

pkg/controller/keystoneendpoint/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,227 @@
...
+// a Pod as an example
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileKeystoneEndpoint) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling KeystoneEndpoint")
+
+	// Fetch the KeystoneEndpoint instance
+	instance := &keystonev1.KeystoneEndpoint{}
```
dprince, line 93: I left the same comment on the KeystoneService controller but we'll want to block here to ensure the actual keystone service is bootstrapped before continuing. Doesn't have to be this spot, but somewhere in the controller.

deploy/crds/keystone.openstack.org_keystoneservices_crd.yaml
```diff
@@ -0,0 +1,42 @@
...
+    listKind: KeystoneServiceList
+    plural: keystoneservices
+    singular: keystoneservice
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: KeystoneService is the Schema for the keystoneservices API
+      properties:
```
bogdando, line 18: it seems the properties are missing here?

deploy/crds/keystone.openstack.org_v1_keystoneservice_cr.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: keystone.openstack.org/v1
+kind: KeystoneService
+metadata:
+  name: example
+spec:
+  serviceName: example
```
bogdando, line 6: ^^ like these properties

pkg/apis/keystone/v1/keystoneendpoint_types.go
```diff
@@ -0,0 +1,59 @@
...
+
+// KeystoneEndpoint is the Schema for the keystoneendpoints API
+// +k8s:openapi-gen=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=keystoneendpoints,scope=Namespaced
+type KeystoneEndpoint struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   KeystoneEndpointSpec   `json:"spec,omitempty"`
```
bogdando, line 44: ditto

deploy/crds/keystone.openstack.org_keystoneservices_crd.yaml
bogdando: nm it's spec/properties/spec ...

pkg/controller/keystoneendpoint/keystoneendpoint_controller.go
bogdando: Why blocking things instead of leaving it for the another reconciliation attempt? I thought blocking things in reconcile is not recommended.

pkg/controller/keystoneendpoint/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	// Fetch the service ID
+	listOpts := services.ListOpts{
+		ServiceType: instance.Spec.ServiceType,
+	}
+	allPages, err := services.List(identityClient, listOpts).AllPages()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	allServices, err := services.ExtractServices(allPages)
+	reqLogger.Info("allServices", "allServices", fmt.Sprintf("%v", allServices))
```
bogdando, line 138: if there was an error, what would be held in allServices? Probably nothing good?

pkg/controller/keystoneservice/keystoneservice_controller.go
```diff
@@ -0,0 +1,163 @@
...
+		service, err := services.Create(identityClient, createOpts).Extract()
+		if err != nil {
+			reqLogger.Error(err, "error")
+			return reconcile.Result{}, err
+		}
+
+		// Set ServiceID in the status
+		reqLogger.Info("instance.Status.ServiceID", "ServiceID", instance.Status.ServiceID)
+		reqLogger.Info("service.ID", "service.ID", service.ID)
+		if instance.Status.ServiceID != service.ID {
```
bogdando, line 138: isn't instance.Status.ServiceID is always "" under this conditional path?

pkg/controller/keystoneendpoint/keystoneendpoint_controller.go
slagle: i think we need to decide on the pattern we want here since this will be a scenario we will see in many places. I tend to agree with bogdan that returning a requeue result seems like it would be more of a k8s like pattern. we can even set a delay before the reconcile is attempted again.

could we set something like reconcileAttempts in the Status, and if we exceed a certain amount, then return an error? That, in combination with the reconcile delay would give us what we need to block sufficiently without leaving the reconcile loop blocking for a long period of time.

dprince: Yes. Result.Requeue is what I was referring too about. That would effectively block the controller from finishing until keystone bootstrapping is completed. I'm already using Result.Requeue in the main keystone controller fwiw, and the pattern is very common across operators

dprince:
Going to merge this as-is. We agreed to follow up on some of the inline comments in a follow up patch soon :)

pkg/controller/keystoneendpoint/keystoneendpoint_controller.go
bogdando: this log info should be moved after the err check

pkg/controller/keystoneservice/keystoneservice_controller.go
bogdando: still unclear to me what updates instance.Status.ServiceID to become not an empty string...

PR #16, opened 2020-07-22
Wait for Keystone bootstrap in endpoint/service controllers

slagle:
The endpoint and service controllers are updated so that they ensure the
keystone boostrap job is complete by checking the bootstrap hash from
KeystoneAPI.Status.

### DIFF
--- a/pkg/controller/keystoneservice/keystoneservice_controller.go
+++ b/pkg/controller/keystoneservice/keystoneservice_controller.go
@@ -2,11 +2,11 @@ package keystoneservice
 
 import (
 	"context"
-
 	gophercloud "github.com/gophercloud/gophercloud"
 	openstack "github.com/gophercloud/gophercloud/openstack"
 	services "github.com/gophercloud/gophercloud/openstack/identity/v3/services"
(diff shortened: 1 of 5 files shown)


pkg/controller/keystoneservice/keystoneservice_controller.go
```diff
@@ -86,9 +87,28 @@ func (r *ReconcileKeystoneService) Reconcile(request reconcile.Request) (reconci
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling KeystoneService")
 
+	keystoneAPI := keystone.API(request.Namespace, "keystone")
```
dprince, line 90: This hard codes the name of the API to keystone though? I think we'll have to search for keystone objects in this namespace right and use the one we find...

dprince: Going to go ahead and merge this. I will add a comment which identifies that identifies this as being hard coded for now.

PR #19, opened 2020-07-24
Olm fixes

dprince:


dprince:
P.S. I've gone and pushed these to quay.io/openstack-k8s-operators/keystone-operator:v0.0.3 as without them it won't work

stuggi:
lgtm

PR #21, opened 2020-07-30
Add a push target for the operator

mdbooth:
Add a target to push to image repository of the logged-in cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,15 +10,59 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-all: csv-generator
+all: operator
 
-csv-generator:
-	CGO_ENABLED=0 go build -a -ldflags '-extldflags "-static"' -o build/_output/csv-generator tools/csv-generator.go
+csv-generator: build/_output/csv-generator
+build/_output/csv-generator: tools/csv-generator.go
+	CGO_ENABLED=0 go build -a -ldflags '-extldflags "-static"' -o $@ $<
+
+operator-image: build/_output/csv-generator build/_registry/image_label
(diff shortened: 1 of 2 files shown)


dprince:
@mpryc would be good to know your thoughts on this. I mentioned today that I felt like this duplicates some of the build logic you have been working on.

Makefile
```diff
@@ -10,15 +10,59 @@
...
+all: operator
 
-csv-generator:
-	CGO_ENABLED=0 go build -a -ldflags '-extldflags "-static"' -o build/_output/csv-generator tools/csv-generator.go
+csv-generator: build/_output/csv-generator
+build/_output/csv-generator: tools/csv-generator.go
+	CGO_ENABLED=0 go build -a -ldflags '-extldflags "-static"' -o $@ $<
+
+operator-image: build/_output/csv-generator build/_registry/image_label
+	operator-sdk build --image-builder buildah $$(cat build/_registry/image_label)
```
mpryc, line 20: I would be in favor of using https://quay.io/repository/openstack-k8s-operators/builder-toolbox to invoke commands such as operator-sdk or buildah.

This way we will have consistent set of tools and with the image tags we can pin or select specific version of those tools independent of host OS.

dprince:
so if we add this to keystone-operator we'd also do similar to other Makefiles as well? I might prefer to just use builders myself and keep the operators* codebases separated from the development environment scripts a bit

mdbooth:
@dprince Yes we'd want this to be common. The point about mixing this into the repo is a good one. I really want the workflow this enables, though, so I might look at making it common in the ansible.

PR #24, opened 2020-08-06
Use MariaDBSchema CRD to create DB schema

dprince:
This updates the keystone controller so that it rely's
on the mariadb-operator's MariaDBSchema to create
databases for us.

Important to note that the implementation uses the
unstructured k8s client so that we don't explicitly
depend on the mariadb-operator code struct's (and associated
golang lib versions).

The old code is removed from the dbsync job and we can
also remove the associated parameters.

### DIFF
--- a/pkg/controller/keystoneapi/keystoneapi_controller.go
+++ b/pkg/controller/keystoneapi/keystoneapi_controller.go
@@ -13,6 +13,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"reflect"
@@ -24,6 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
 	"strings"
 	"time"
 )
@@ -173,6 +175,30 @@ func (r *ReconcileKeystoneAPI) Reconcile(request reconcile.Request) (reconcile.R
 			return reconcile.Result{}, err
 		}
 		return reconcile.Result{RequeueAfter: time.Second * 5}, err
+	}
+
+	// Create the DB Schema (unstructured so we don't explicitly import mariadb-operator code)
+	schemaObj, err := keystone.SchemaObject(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	foundSchema := &unstructured.Unstructured{}
+	foundSchema.SetGroupVersionKind(schemaObj.GroupVersionKind())
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: schemaObj.GetName(), Namespace: schemaObj.GetNamespace()}, foundSchema)
+	if err != nil && k8s_errors.IsNotFound(err) {
+		err := r.client.Create(context.TODO(), &schemaObj)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	} else if err != nil {
+		return reconcile.Result{}, err
+	} else {
+		completed, _, err := unstructured.NestedBool(foundSchema.UnstructuredContent(), "status", "completed")
+		if !completed {
+			reqLogger.Info("Waiting on DB to be created...")
+			return reconcile.Result{RequeueAfter: time.Second * 5}, err
+		}
 	}
 
 	// Define a new Job object
--- a/pkg/keystone/schema.go
+++ b/pkg/keystone/schema.go
@@ -1,0 +1,27 @@
+package keystone
+
+import (
+	comv1 "github.com/openstack-k8s-operators/keystone-operator/pkg/apis/keystone/v1"
+	util "github.com/openstack-k8s-operators/lib-common/pkg/util"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"strings"
+)
+
+type schemaOptions struct {
+	DatabaseHostname string
+	SchemaName       string
+	SchemaPassword   string
+}
+
+// SchemaObject func
+func SchemaObject(cr *comv1.KeystoneAPI) (unstructured.Unstructured, error) {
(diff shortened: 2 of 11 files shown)


mdbooth:
I don't think the MariaDB CRD is baked enough for this, yet. It needs to be updated, and that's going to be much harder if we already have dependencies on it. I'd wait on this one.

dprince:
@mdbooth we spoke on IRC and you mentioned your concern here is related to the use of secrets. Do we have agreement on how we are implementing those across the board. When I checked nova-operator uses a single "osp-secrets" for all services. It appears we are using individual secrets per service for the dev-tools YAML's.

Why not land this and loop back and fine tune the secrets later across the board? Yes that is a breaking change but what I have here is already so much better in that it doesn't require the *Maria DB Admin creds* in the keystone CR's anymore right?

mdbooth:
> @mdbooth we spoke on IRC and you mentioned your concern here is related to the use of secrets. Do we have agreement on how we are implementing those across the board. When I checked nova-operator uses a single "osp-secrets" for all services. It appears we are using individual secrets per service for the dev-tools YAML's.

I don't (yet) have a well-formed opinion on this beyond: Secrets good, CRs bad. My very malleable current inclination would be towards single-credential secrets.

pkg/keystone/schema.go
```diff
@@ -0,0 +1,35 @@
+package keystone
```
mdbooth, line 1: We're going to have to create this exact same package everywhere we use it. I wouldn't be a fan of that, and hence using Unstructured for this purpose.

However, I have also been wondering if we should rename the Keystone operator to the ControlPlane operator, and add all the control plane CRDs, with their own controllers, to the same operator. We could move it to lib-common, but that would also suck more than just importing it from mariadb-operator directly.

pkg/keystone/schema.go
```diff
@@ -0,0 +1,35 @@
...
+
+// SchemaObject func
+func SchemaObject(cr *comv1.KeystoneAPI) unstructured.Unstructured {
+	opts := schemaOptions{cr.Spec.DatabaseHostname, cr.Name, cr.Spec.DatabasePassword}
+
+	decoder := yaml.NewYAMLOrJSONDecoder(strings.NewReader(util.ExecuteTemplateFile("mariadb_schema.yaml", &opts)), 4096)
+	u := unstructured.Unstructured{}
+	err := decoder.Decode(&u)
+	if err == io.EOF {
+		panic(err.Error())
```
mdbooth, line 27: Maybe, because you could argue that this is unrecoverable and an error in the operator image. My instinct is still probably not, though.

pkg/keystone/schema.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	if err != nil {
+		panic(err.Error())
```
mdbooth, line 30: Definitely no.  This error can be caused by user input (because we're not checking escaping in the yaml.

Incidentally, is that a security issue? I would assume yes until proven otherwise.

pkg/keystone/schema.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+type schemaOptions struct {
+	DatabaseHostname string
+	SchemaName       string
+	SchemaPassword   string
+}
+
+// SchemaObject func
+func SchemaObject(cr *comv1.KeystoneAPI) unstructured.Unstructured {
```
mdbooth, line 20: This function can generate errors, so its return **must** include an error. Panic isn't acceptable. We don't want to be trying to unwind these later. Lets get this right first time.

pkg/controller/keystoneapi/keystoneapi_controller.go
```diff
@@ -175,6 +177,27 @@ func (r *ReconcileKeystoneAPI) Reconcile(request reconcile.Request) (reconcile.R
...
 	}
 
+	// Create the DB Schema (unstructured so we don't explicitly import mariadb-operator code)
+	schemaObj := keystone.SchemaObject(instance)
+
+	foundSchema := &unstructured.Unstructured{}
+	foundSchema.SetGroupVersionKind(schemaObj.GroupVersionKind())
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: schemaObj.GetName(), Namespace: schemaObj.GetNamespace()}, foundSchema)
+	if err != nil && k8s_errors.IsNotFound(err) {
+		err := r.client.Create(context.TODO(), &schemaObj)
```
mdbooth, line 187: I initially worried that this could be a race, but I think not because all created objects should have the same name, so only the first should succeed.

Incidentally, although we don't (currently, anyway) intend for this code to runnable concurrently, I would defensively always assume that it might run concurrently.

pkg/keystone/schema.go
dprince: sigh, I thought I removed these. I will fix these up no problem

pkg/keystone/schema.go
dprince: We do not want this type of code in lib-common as it would significantly add to the set of goland dependencies we have to keep in lockstep. I think the idea with go libraries is keep dependencies simple. What we could do is have a lib-copy where we have tests on common code that we (manually) copy into some of our operators.

With regards to renaming KeystoneOperator I would ask this: Is there any chance someone might want to use parts of our Ctlplane elsewhere? Also putting all apps into the same repos seems to go against best practices (https://github.com/operator-framework/community-operators/blob/master/docs/best-practices.md) "If an application consists of multiple different tiers or components, multiple Operators should be written for each of them. If the application for example consists of Redis, AMQ and MySQL, there should be 3 Operators, not one."

dprince: The lib-copy idea I mentioned above is in fact how oslo in OpenStack first started out btw. (we copied things around). It wasn't great, but that could provide re-use albeit via periodic (manual) copying of the libs around. And we could have a CI job alert of us things are out of sync.

pkg/controller/keystoneapi/keystoneapi_controller.go
dprince: only 1 Reconcile should run at a time. That said I think the code as written is probably robust enough to handle the case of concurrency in that if 2 threads somehow ran at the same time they would compete and 1 would pretty quickly throw a Reconcile Error and thus retry again. Essentially the way we are writing operators is always in a "make it so" fashion. If there is an error you log it, and retry accordingly hopefully eventually getting to the state you wanted.

dprince:
I've now changed the MariaDBSchema to use secrets. The implementation looks like this:

https://github.com/openstack-k8s-operators/mariadb-operator/blob/master/config/samples/database_v1beta1_mariadbschema.yaml

Until I updates this PR please don't merge.

dprince:
@mdbooth thanks a lot for your feedback here. In rebasing this I've pushed a new PR which incorporates both this patch along with Keystone secrets updates as well. Please have a look: https://github.com/openstack-k8s-operators/keystone-operator/pull/27

PR #26, opened 2020-08-18
Reworked Dockerfile.rhel for new pattern

yazug:
with reworked operator-sdk the Dockerfile pattern shifted.  Updated
Dockerfile.rhel for the new pattern to see if we can get closer to
a single templated version that can be used in either system with less
changes.

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -1,0 +1,62 @@
+# golang-builder is used in OSBS build
+ARG GOLANG_BUILDER=openshift/golang-builder:1.13
+ARG OPERATR_BASE_IMAGE=ubi8-minimal:8.1-released
+
+FROM ${GOLANG_BUILDER} AS builder
+
+# Intended to build in OSBS using cachito external sources bundle
+ARG REMOTE_SOURCE=.
+ARG REMOTE_SOURCE_DIR
+ARG REMOTE_SOURCE_SUBDIR=app
+ARG DEST_ROOT=/dest-root
+
+COPY $REMOTE_SOURCE $REMOTE_SOURCE_DIR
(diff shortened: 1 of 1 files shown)


Dockerfile.rhel
```diff
@@ -0,0 +1,62 @@
...
+
+# install our templates
+RUN  mkdir -p ${OPERATOR_TEMPLATES}
+COPY --from=builder $DEST_ROOT/templates ${OPERATOR_TEMPLATES}
+
+# install CRDs and required roles, services, etc
+RUN  mkdir -p ${OPERATOR_BUNDLE}
+
+COPY --from=builder $DEST_ROOT/config/crd/bases/keystone.openstack.org_keystoneapis.yaml ${OPERATOR_BUNDLE}/keystone.openstack.org_keystoneapis_crd.yaml
+COPY --from=builder $DEST_ROOT/config/crd/bases/keystone.openstack.org_keystoneendpoints.yaml ${OPERATOR_BUNDLE}/keystone.openstack.org_keystoneendpoints_crd.yaml
```
yazug, line 50: it might be nice to move this part with the renaming and mangling to the builder part so that we can just copy the folder over in this step

mpryc:
Nice.

Dockerfile.rhel
yazug: https://github.com/yazug/keystone-operator/commit/a5d135f9b8bd9bf406b6e660e7dd812a529efdfa was the idea I was thinking of for cleaning up the crd renaming and sed work

yazug: I have branch with several patches rotating the Dockerfile and the Dockerfile.rhel to be closer to the same https://github.com/yazug/keystone-operator/tree/dockerfile_rework

yazug:
abandoning this in favour or PR #29

PR #27, opened 2020-08-18
Secrets and use MariaDBSchema CRD

dprince:
Updates the keystone CRD to use secrets.

Additionally, this updates the keystone controller so that it rely's
on the mariadb-operator's MariaDBSchema to create databases for us.

Important to note that the implementation uses the unstructured k8s
client so that we don't explicitly depend on the mariadb-operator code
struct's (and associated golang lib versions).

The old code is removed from the dbsync job and we can also remove
the associated parameters.

### DIFF
--- a/pkg/keystone/bootstrap.go
+++ b/pkg/keystone/bootstrap.go
@@ -10,16 +10,15 @@ import (
 )
 
 type bootstrapOptions struct {
-	AdminPassword string
-	APIEndpoint   string
-	ServiceName   string
+	APIEndpoint string
+	ServiceName string
 }
 
 // BootstrapJob func
 func BootstrapJob(cr *keystonev1beta1.KeystoneAPI, configMapName string, APIEndpoint string) *batchv1.Job {
 
 	// NOTE: as a convention the configmap is name the same as the service
-	opts := bootstrapOptions{cr.Spec.AdminPassword, APIEndpoint, configMapName}
+	opts := bootstrapOptions{APIEndpoint, configMapName}
 	runAsUser := int64(0)
 
 	job := &batchv1.Job{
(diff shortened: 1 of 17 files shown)


pkg/keystone/volumes.go
```diff
@@ -79,6 +100,44 @@ func getVolumeMounts() []corev1.VolumeMount {
...
+		{
+			MountPath: "/var/lib/emptydir",
+			ReadOnly:  false,
+			Name:      "emptydir",
+		},
 	}
 
 }
+
+// common Keystone API VolumeMounts (db-kolla-config mounts passwords directly as keystone-api.conf)
```
mdbooth, line 112: keystone-api.conf is weird keystone-passwords.conf?

pkg/keystone/bootstrap.go
```diff
@@ -32,6 +31,51 @@ func BootstrapJob(cr *keystonev1beta1.KeystoneAPI, configMapName string, APIEndp
 				Spec: corev1.PodSpec{
 					RestartPolicy:      "OnFailure",
 					ServiceAccountName: "keystone",
+
+					InitContainers: []corev1.Container{
```
mdbooth, line 35: It looks like all 3 uses of this init container are identical. This should be common code.

templates/mariadb_schema.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: database.openstack.org/v1beta1
+kind: MariaDBSchema
+metadata:
+  name: {{.SchemaName}}
+  labels:
+    dbName: {{.DatabaseHostname}}
```
mdbooth, line 6: I haven't looked hard at MariaDB&lt;Schema&gt; but is this really how we're specifying the database server?

mdbooth (changes requested):
This patch would significantly add to our technical debt. We need to get the MariaDBSchema rename out of the way so we don't continue to proliferate the word 'schema' into more places.

Also, seeing how it's used makes me think we need changes in that object beyond its name.

pkg/keystone/bootstrap.go
```diff
@@ -32,6 +31,51 @@ func BootstrapJob(cr *keystonev1beta1.KeystoneAPI, configMapName string, APIEndp
...
+						{
+							Name:    "keystone-secrets",
+							Image:   cr.Spec.ContainerImage,
+							Command: []string{"/bin/sh", "-c", util.ExecuteTemplateFile("password_init.sh", nil)},
+							Env: []corev1.EnvVar{
+								{
+									Name:  "DatabaseHost",
```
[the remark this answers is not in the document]
dprince: The reason for camelcasing them is solely because that matches the secret entries and thus makes grep'ability greater.

dprince:
We can agree to disagree but I'm going to go ahead and land this for the secrets and keystone CRD simplifications alone.

I have created a Jira issue where we are tracking the MariaDBSchema name changes (where we'll rename MariaDBSchema to MariaDBDatabase). This will involve changes to mariadb-operator, glance-operator, and keystone-operator.

PR #29, opened 2020-08-19
Dockerfile rework

yazug:
This pull request will update the Dockerfile to use arguments as well as introduce a Dockerfile.rhel which is only different via the ARG's passed in for building

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -1,14 +1,15 @@
 # golang-builder is used in OSBS build
-ARG GOLANG_BUILDER=golang:1.13
-ARG OPERATOR_BASE_IMAGE=registry.access.redhat.com/ubi7/ubi-minimal:latest
+ARG GOLANG_BUILDER=openshift/golang-builder:1.13
+ARG OPERATOR_BASE_IMAGE=registry.redhat.io/ubi8/ubi-minimal:latest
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,49 +1,65 @@
...
+RUN cp config/crd/bases/keystone.openstack.org_keystoneendpoints.yaml ${DEST_ROOT}/bundle/keystone.openstack.org_keystoneendpoints_crd.yaml
+RUN cp config/crd/bases/keystone.openstack.org_keystoneservices.yaml ${DEST_ROOT}/bundle/keystone.openstack.org_keystoneservices_crd.yaml
+
+# strip top 2 lines (this resolves parsing in opm which handles this badly)
+RUN sed -i -e 1,2d ${DEST_ROOT}/bundle/*
+
+FROM ${OPERATR_BASE_IMAGE}
+ARG DEST_ROOT=/dest-root
+
+LABEL   com.redhat.component="keystone-operator-container" \
```
mpryc, line 38: I wonder if those can be args as well? I mean like all per operator keep as vars, this way any new operator will have exactly same bottom part of Dockerfile and only vary with beginning ?

yazug: could, but let's get incrementally closer between upstream and downstream and then look at across the different operators

If I recall correctly there are currently 3 or 4 variants

PR #46, opened 2020-10-19
Add username and password to KeystoneService CRD

slagle:


dprince (approved):

controllers/keystoneservice_controller.go
```diff
@@ -204,7 +204,23 @@ func (r *KeystoneServiceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
-	err = reconcileUser(reqLogger, identityClient, instance.Spec.ServiceName)
+	var username string
+	if instance.Spec.Username == "" {
+		username = instance.Spec.ServiceName
+	} else {
+		username = instance.Spec.Username
+	}
+
+	var password string
+	if instance.Spec.Password == "" {
```
stuggi, line 215: shouldn't this be:
`if instance.Spec.Password != ""`

stuggi (changes requested):

controllers/keystoneservice_controller.go
```diff
@@ -330,6 +346,7 @@ func reconcileUser(reqLogger logr.Logger, client *gophercloud.ServiceClient, use
 		createOpts := users.CreateOpts{
 			Name:             username,
 			DefaultProjectID: serviceProjectID,
+			Password:         password,
 		}
 		user, err := users.Create(client, createOpts).Extract()
 		if err != nil {
```
stuggi, line 352: not important atm, but if pass for two services the same user we probably need to catch it here to not get an error and just ignore it, or check first if the user is already there and run an update instead of create.

controllers/keystoneservice_controller.go
slagle: yea, fixed that

stuggi (approved):

PR #53, opened 2022-05-13
Set uniqe instance sample name

stuggi:
The instance name is used to create the DB for the service,
lets use a uniqe name for the keystoneapi for the sample to
not mess with other services sample.

### DIFF
--- a/config/samples/keystone_v1beta1_keystoneapi.yaml
+++ b/config/samples/keystone_v1beta1_keystoneapi.yaml
@@ -1,7 +1,7 @@
 apiVersion: keystone.openstack.org/v1beta1
 kind: KeystoneAPI
 metadata:
-  name: openstack
+  name: keystone
 spec:
   containerImage: quay.io/tripleotraincentos8/centos-binary-keystone:current-tripleo
   replicas: 1


dprince:
I assume this is because we had previously had some k8s resources named directly after the CR name like Services, Databasename, etc. I was planning on adding a suffix (or prefix) to the created resource names such that we avoid collisions here altogether. So for example the service for keystone would be 'openstack-keystone' or 'foo-keystone' where openstack or foo would be the respective CR name that triggers the creation of resouces with those names. @stuggi thoughts?

dprince:
@stuggi something like this instead: https://github.com/openstack-k8s-operators/keystone-operator/pull/54

config/samples/keystone_v1beta1_keystoneapi.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: keystone.openstack.org/v1beta1
 kind: KeystoneAPI
 metadata:
-  name: openstack
+  name: keystoneapi
```
dprince, line 4: Can we just make it keystone?

PR #55, opened 2022-05-27
This change add runAsUser to the security context to avoid the following error:

SeanMooney:
container has runAsNonRoot and image has non-numeric user (nonroot),
cannot verify user is non-root

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,8 @@ RUN cp -r templates ${DEST_ROOT}/templates
 FROM $OPERATOR_BASE_IMAGE
 
 ARG DEST_ROOT=/dest-root
-ARG USER_ID=nonroot:nonroot
+# NONROOT default id https://github.com/GoogleContainerTools/distroless/blob/main/base/base.bzl#L8=
+ARG USER_ID=65532
 
 ARG IMAGE_COMPONENT="keystone-operator-container"
 ARG IMAGE_NAME="keystone-operator"
@@ -56,7 +57,7 @@ LABEL com.redhat.component="${IMAGE_COMPONENT}" \
       io.openshift.tags="${IMAGE_TAGS}"
 ### DO NOT EDIT LINES ABOVE
 
-ENV USER_UID=1001 \
+ENV USER_UID=$USER_ID \
     OPERATOR_TEMPLATES=/usr/share/keystone-operator/templates/ \
     WATCH_NAMESPACE=openstack,openshift-machine-api,openshift-sriov-network-operator
 


dprince:
@SeanMooney just curious what container image you are using to hit the error you mention?

SeanMooney:
by the way I'm pretty sure the reason this is happing is because the docker file is incorrect

its taking the user id as a build are
and defaulting to nonroot:nonroot
ARG USER_ID=nonroot:nonroot
https://github.com/openstack-k8s-operators/keystone-operator/blob/master/Dockerfile#L36


then later you are exporting a USER_ID env var
ENV USER_UID=1001 
https://github.com/openstack-k8s-operators/keystone-operator/blob/master/Dockerfile#L59

and then setting the user 
USER $USER_ID
https://github.com/openstack-k8s-operators/keystone-operator/blob/master/Dockerfile#L71

so currently the docker file sets USER to the build arg USER_ID
not the env var USER_UID

so the other way to fix this would be to use
USER $USER_UID on line 71

that way the runAsNonRoot validation would pass as its numeric and not 0

env directives have precedence over build args as can be seen with a simple test case


sean@cloud:~/temp$ cat Containerfile 
FROM busybox:latest
ARG VAR_1=1
ENV VAR_1=42
RUN echo "print variable value:" $VAR_1
sean@cloud:~/temp$ podman build --build-arg VAR_1=test -t test .
STEP 1/4: FROM busybox:latest
STEP 2/4: ARG VAR_1=1
--> 1040597d28c
STEP 3/4: ENV VAR_1=42
--> c1c1b56df94
STEP 4/4: RUN echo "print variable value:" $VAR_1
print variable value: 42
COMMIT test
--> 36e8aea92de
Successfully tagged localhost/test:latest
36e8aea92debb86c438487191eb46dc83e5e4011508d007e4417a058b3898178

so we likely should have one or the other but the current USER_ID=nonroot:nonroot build arg default will not work with runAsNonRoot so we either need to set a user in the container spec or we need to change it to be a numeric value in the docker file

dprince:
@SeanMooney This PR is basically the same as your MariaDB PR. Since you have more comments there perhaps we could just continue the conversation there... and once we agree on any needed fixes we could then apply them to all the operators? (trying to minimize the review time, etc for folks)

SeanMooney:
yep we can continue with which ever one is simpler for you to work with.

I don't really have a preference.

SeanMooney:
this minimal change is enough to resolve the issue i had.
if you like this approch i can update the mariadb pr too

dprince (approved):

dprince:
> this minimal change is enough to resolve the issue i had. if you like this approch i can update the mariadb pr too

This seems okay. Lets do it in mariadb as well. Thanks @SeanMooney

PR #56, opened 2022-06-09
Update keystone-operator to use common functions

stuggi:
To make different service operators can look similar
this moves implementation of the different required
resource to lib-common with interfaces that they
could be consumed easily. This demonstrates how it
could be done.

Also:
- service customization via CustomServiceConfig param
which gets added to the VM and inside the pod as
keystone/conf.d/custom.conf
- Nodeselector
- pod affinity - if possible two pods of the same
service should not run on the same worker node. If this
is not possible	the get still created on the same worker
node.

TODOs:
- finish conditions, right now only in db create
- set finalizers on the different created resources
to prevent deletion
- poc update to see if the dedicated update reconcile
func is required
- poc upgrade
- implement DefaultConfigOverwrite
- move clouds.yaml generate to a openstack upper operator
- ...

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/15


stuggi:
needs go.mod update when https://github.com/openstack-k8s-operators/lib-common/pull/15 merged

go.mod
```diff
@@ -75,3 +76,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+// 1. push your code to github.com/fork/somelib @ dev branch
+// 2. modify your go.mod file, add a line replace github.com/original/somelib => github.com/fork/somelib dev
+// 3. execute go mod tidy command. After done these, go will auto replace the dev in go.mod to a suitiable pseudo-version.
+replace github.com/openstack-k8s-operators/lib-common => github.com/stuggi/lib-common v0.0.0-20220609154431-ce6d282845fb
```
stuggi, line 83: need to be removed when dependency on lib-common merged

stuggi:
can be tested using install_yamls with `KEYSTONE_IMG=quay.io/mschuppe/keystone-operator-index:v0.0.36 KEYSTONE_REPO=https://github.com/stuggi/keystone-operator.git KEYSTONE_BRANCH=rework make keystone_deploy`

pkg/keystone/const.go
```diff
@@ -0,0 +1,56 @@
...
+	KeystoneInternalPort int32 = 5000
+
+	// AppSelector -
+	AppSelector = "app"
+
+	// KollaConfig -
+	KollaConfig = "/var/lib/config-data/merged/keystone-api-config.json"
+
+	// InitContainerCommand -
+	InitContainerCommand = "/usr/local/bin/container-scripts/init.sh"
```
dprince, line 47: Are constants like these really needed though? Unless these commands are used in more than 1 place I sort of thing having them in const.go instead of the resulting packages makes them harder to follow.

stuggi: the idea was, when apply this to the other operators, its probably good to just have to do customization in a single place and not in several files. But sure, we could migrate those which are only used once to the place where its used.

pkg/keystone/const.go
```diff
@@ -0,0 +1,56 @@
...
+	// DBSyncCommand -
+	DBSyncCommand = "/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start"
```
SeanMooney, line 49: should this not be keystone-manage db sync instead of koala_start

SeanMooney: I guess you are relying on https://github.com/openstack/tripleo-common/blob/master/container-images/kolla/keystone/extend_start.sh#L25

but its probably better to do the db sync directly

templates/keystoneapi/config/keystone-api-config.json
```diff
@@ -2,21 +2,27 @@
...
             "dest": "/etc/httpd/conf/httpd.conf",
             "owner": "root",
             "perm": "0644"
         },
         {
-            "source": "/var/lib/config-data/logging.conf",
+            "source": "/var/lib/config-data/merged/logging.conf",
             "dest": "/etc/keystone/logging.conf",
-            "owner": "keystone",
+            "owner": "root",
```
SeanMooney, line 25: why root?  all of the config should be owned by the keystone user we shouldn't be running the keystone service as root.

templates/keystoneapi/config/keystone-api-config.json
```diff
@@ -2,21 +2,27 @@
...
-            "source": "/var/lib/config-data/httpd.conf",
+            "source": "/var/lib/config-data/merged/custom.conf",
+            "dest": "/etc/keystone/keystone.conf.d/custom.conf",
+            "owner": "keystone",
+            "perm": "0600"
+        },
+        {
+            "source": "/var/lib/config-data/merged/httpd.conf",
             "dest": "/etc/httpd/conf/httpd.conf",
             "owner": "root",
```
SeanMooney, line 19: this is presumable because apache is running as root? can we avoid that somehow.

config/rbac/role.yaml
```diff
@@ -8,12 +8,13 @@ rules:
 - apiGroups:
   - apps
   resources:
-  - deployments
+  - statefulsets
```
SeanMooney, line 11: so correct me if I'm wrong but keystone does not store any state in the container or in persistent volumes.

all its state is stored in the db correct? has that changed with fernet tokens? we should avoid stateful sets for no stateful services so I want to understand if you are making this a statefulset because you need each instance of keystone to use a sepert set of pvc or not.

pkg/keystone/statefulset.go
```diff
@@ -0,0 +1,108 @@
...
+	keystonev1beta1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
+	common "github.com/openstack-k8s-operators/lib-common/pkg/common"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StatefulSet func
+func StatefulSet(
```
SeanMooney, line 28: looking at the volume mounts in below the fernet keys are mounted readonly and I don't see any other volumes that would require this to be a statefulset instead of a deployment so can you either document in a comment why this need to be a stateful set or revert this to a deployment?

config/rbac/role.yaml
stuggi: its only because I like having the pods named keystone-0, -1 ,-2 instead of the deployment naming keystone-xyzpb. thats the only reason. That would make it easier to see the number of running instances. we can switch it back to be a deployment.

pkg/keystone/const.go
stuggi: yes correct right now I depend on the kolla scripts

pkg/keystone/statefulset.go
stuggi: see above, just the naming for listing services. I'll switch it back if we do not want it.

templates/keystoneapi/config/keystone-api-config.json
stuggi: because keystone only need to be able to read it. so no need to be the owner of logging.conf

stuggi:
will switch back to use a deployment

pkg/keystone/statefulset.go
stuggi: moved back to use a deployment

config/rbac/role.yaml
stuggi: moved back to use a deployment

pkg/keystone/jobs.go
```diff
@@ -0,0 +1,185 @@
...
+	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
+	keystonev1beta1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
+
+	common "github.com/openstack-k8s-operators/lib-common/pkg/common"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// BootstrapJob func
```
dprince, line 28: Moving these functions into jobs.go in a follow up would make the changes here a lot easier to review. Essentially keep BootstrapJob in bootstrap.go and DbSyncJob in dbsync.go. I'd be fine to eventually move them but with patches like this minimizing structural moves like this is beneficial I think.

templates/keystoneapi/config/keystone-api-config.json
dprince: The container itself isn't being granted privs on the cluster so it should be fine right? This has more to do with the defaults of Apache service itself so we should probable stick with root I think to avoid other issues.

templates/keystoneapi/config/keystone-api-config.json
dprince: yeah, by default Apache runs as root

pkg/keystone/const.go
dprince: I think depending on the KOLLA_SCRIPTS is probably fine. If we hit issues we could always switch to inline later too

pkg/keystone/const.go
dprince: Its just that having the constants here makes the other files/jobs/etc hard to follow. I'm all for constants where they are shared and you want one place to edit them. I'm generally a fan of having source code be more readable and in this case a constant at the top of the actual file these are used in would better suit

pkg/keystone/jobs.go
stuggi: done

PR #60, opened 2022-06-21
Change order to bootstrap service before creating deployment

stuggi:
The operator is running the keystone bootstrap job after the
deployment is up. Because there is also no readiness probe,
this means it's not possible for a client to know if it's supposed
to be working, yet.

This changes the order to expose the service (get the endpoint
URLs) and rund the bootstrap job before the deployment of the
keystone service gets created.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -176,7 +176,12 @@ func (r *KeystoneAPIReconciler) reconcileDelete(ctx context.Context, instance *k
 	return ctrl.Result{}, nil
 }
 
-func (r *KeystoneAPIReconciler) reconcileInit(ctx context.Context, instance *keystonev1.KeystoneAPI, helper *helper.Helper) (ctrl.Result, error) {
+func (r *KeystoneAPIReconciler) reconcileInit(
+	ctx context.Context,
+	instance *keystonev1.KeystoneAPI,
+	helper *helper.Helper,
+	endpointLabels map[string]string,
+) (ctrl.Result, error) {
 	r.Log.Info("Reconciling Service init")
 
 	//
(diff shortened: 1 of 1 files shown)


controllers/keystoneapi_controller.go
```diff
@@ -250,6 +255,72 @@ func (r *KeystoneAPIReconciler) reconcileInit(ctx context.Context, instance *key
...
+	ctrlResult, err = bootstrapjob.DoJob(
+		ctx,
+		helper,
+	)
+	if (ctrlResult != ctrl.Result{}) {
+		return ctrlResult, nil
+	}
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
abays, line 311: Should we use the same pattern here that we've used elsewhere, where we check for `err != nil` first?  I don't know if it really matters, but it's just something I noticed

controllers/keystoneapi_controller.go
```diff
@@ -250,6 +255,72 @@ func (r *KeystoneAPIReconciler) reconcileInit(ctx context.Context, instance *key
...
+	if bootstrapjob.HasChanged() {
+		instance.Status.Hash[keystonev1.BootstrapHash] = bootstrapjob.GetHash()
+		if err := r.Client.Status().Update(ctx, instance); err != nil {
+			return ctrl.Result{}, err
+		}
+		r.Log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.Hash[keystonev1.BootstrapHash]))
```
abays, line 317: You may have explained this before, but is there a reason that we don't rely on the `defer` status-update approach here and instead directly update it in-flight?

controllers/keystoneapi_controller.go
stuggi: yes I think we could change it again. how the function was implemented to wait for the job to finish required this order to not print errors when e.g. the job was not yet created/available. but I changed this particular case. I'll retest/recheck it and may change it in a follow up.

controllers/keystoneapi_controller.go
stuggi: In general yes we could, but since the hashes are important information to not rerun a job I wanted to save it early when the hash is available and fail early if for some reason we can not update the object before proceed with further steps.

abays (approved):
/lgtm

PR #61, opened 2022-06-21
Allow customize service container resources

stuggi:
Compute Resources required by this service (Limits/Requests).
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/


abays (approved):
/lgtm

PR #64, opened 2022-06-29
Add IsReady method and PasswordSelector

stuggi:
This allows clients to check if the keystoneAPI is up and running
to serve requests. Right now it returns trues if one replica of
the keystoneapi deployement is up. Could be improved later if needed.

Also adds a PasswordSelector which allows to customize which entry
of a secret holds the DB and admin password. Defaults to
KeystoneDatabasePassword and AdminPassord.

Note: KeystoneDatabasePassword is not yet used as the mariadb-operator
does not support passing specific service DB password.


controllers/keystoneapi_controller.go
```diff
@@ -605,7 +606,7 @@ func (r *KeystoneAPIReconciler) reconcileConfigMap(ctx context.Context, instance
 
 	secretName := "openstack-config-secret"
 	var openStackConfigSecret keystone.OpenStackConfigSecret
-	openStackConfigSecret.Clouds.Default.Auth.Password = string(keystoneSecret.Data["AdminPassword"])
+	openStackConfigSecret.Clouds.Default.Auth.Password = instance.Spec.PasswordSelectors.Admin
```
abays, line 609: Should this be `string(keystoneSecret.Data[instance.Spec.PasswordSelectors.Admin])` since `instance.Spec.PasswordSelectors.Admin` holds the selector (the key) for the password and not the value of the password itself?

stuggi: yes it should

abays (approved):
/lgtm

PR #66, opened 2022-06-29
GetKeystoneAPI to return NotFound if keystoneList == 0

stuggi:
If there is no KeystoneAPI found in the namespace and used
labelSelector, return NotFound error. Otherwise the return
statement at the end of the func will fail with pointer
exception.


abays (approved):
/lgtm

PR #69, opened 2022-07-06
update MakeFile for 1.18 support

SeanMooney:
This change updates the MakeFile based on a blank
operator template generated with the latest sdk version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ endif
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.23
+ENVTEST_K8S_VERSION = 1.24.1
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -231,38 +234,3 @@ catalog-build: opm ## Build a catalog image.
...
-	if [ -d  "$(CI_TOOLS_REPO_DIR)" ]; then \
-		echo "Ci tools exists"; \
-		pushd "$(CI_TOOLS_REPO_DIR)"; \
-		git pull --rebase; \
-		popd; \
-	else \
-		git clone $(CI_TOOLS_REPO) "$(CI_TOOLS_REPO_DIR)"; \
-	fi
-
-# Run go fmt against code
```
SeanMooney, line 250: we proablly want to keep this since we added them for ci i guess
ill revert this part

dprince:
For 1.18 support we'll need/want to do much more than this I think and follow operator-sdk upgrade instructions as well.

Makefile
```diff
@@ -147,34 +147,37 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
...
+	mkdir -p $(LOCALBIN)
+
+## Tool Binaries
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+
+## Tool Versions
+KUSTOMIZE_VERSION ?= v3.8.7
+CONTROLLER_TOOLS_VERSION ?= v0.9.0
```
dprince, line 164: The top of this Makefile is rather "cookie cutter" from the SDK. Are the changes you added here based on what we'll end up needing in all operator repos to support 1.18?

Makefile
```diff
@@ -213,7 +216,7 @@ endif
 BUNDLE_IMGS ?= $(BUNDLE_IMG)
 
 # The image tag given to the resulting catalog image (e.g. make catalog-build CATALOG_IMG=example.com/operator-catalog:v0.2.0).
-CATALOG_IMG ?= $(IMAGE_TAG_BASE)-index:v$(VERSION)
+CATALOG_IMG ?= $(IMAGE_TAG_BASE)-catalog:v$(VERSION)
```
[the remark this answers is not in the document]
SeanMooney: well the new default is catalog i can revert this but this was the name used by the SDK for a new repo.

so either we need to maintain index for ever or we can use the default value that is use in the SDK now before we have any users of this.

if we defer changing it and need to change it later it will be more work then changing it now but i don't really mind either way.

Makefile
```diff
@@ -233,6 +236,8 @@ catalog-push: ## Push a catalog image.
 	$(MAKE) docker-push IMG=$(CATALOG_IMG)
 
 
+
+#==========================local ci toolling not form template=============================
```
SeanMooney, line 240: this is the only change that i did manually

i also filtered out some local changes we had such as using podman instead of docker to only include the content that was relevnet

stuggi:
closing this PR as we merged https://github.com/openstack-k8s-operators/keystone-operator/pull/73

PR #70, opened 2022-07-07
Update keystoneservice

stuggi:
* This changes the keystoneservice CRD to remove the plane text password from the spec. To get the pwd like in the service operator pass the Secret and PasswordSelector. Also handles removal of endpoints, service and user of the service on delete using on keystoneservice object delete using finalizer.
    
* Adds KeystoneService type and NewKeystoneService to create a new type. Methods CreateOrPatch to create/patch a KeystoneService and GetServiceID to get the ID of the registerd OpenStack service. Also Delete to delete a KeystoneService.

* Update GetAdminServiceClient to use openstack.NewOpenStack

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/22


stuggi:
TODO needs go.mod update when https://github.com/openstack-k8s-operators/lib-common/pull/22 merged

go.mod
```diff
@@ -3,17 +3,17 @@ module github.com/openstack-k8s-operators/keystone-operator
...
 	github.com/openstack-k8s-operators/lib-common v0.0.0-20220630111354-9f8383d4a2ea
+	github.com/openstack-k8s-operators/lib-common/pkg/openstack v0.0.0-00010101000000-000000000000
 	github.com/openstack-k8s-operators/mariadb-operator v0.0.0-20220516121356-119f8d825a71
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.23.6
-	k8s.io/apimachinery v0.23.6
-	k8s.io/client-go v0.23.6
+	k8s.io/api v0.24.2
+	k8s.io/apimachinery v0.24.2
+	k8s.io/client-go v0.24.2
```
dprince, line 16: Did you mean to bump the k8s API packages in this changeset? Seems unrelated.

controllers/keystoneservice_controller.go
```diff
@@ -101,305 +88,379 @@ func (r *KeystoneServiceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+func (r *KeystoneServiceReconciler) reconcileNormal(
+	ctx context.Context,
+	instance *keystonev1.KeystoneService,
+	helper *helper.Helper,
+	os *openstack.OpenStack,
+) (ctrl.Result, error) {
+	r.Log.Info("Reconciling Service")
+
+	// If the service object doesn't have our finalizer, add it.
+	controllerutil.AddFinalizer(instance, helper.GetFinalizer())
```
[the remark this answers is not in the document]
stuggi: yes correct. when we create it via KeystoneService.CreateOrPatch the end result is that you get 2 finalizers, one from the requesting service and the second form keystoneservice, but those get removed in the chain when deleted

go.mod
stuggi: ah no, not intentionally. I think the root cause is when I created the openstack module it got the 0.24.2 version and due to that import it also changed it here. let me return it there to 0.23.6 and we can update it later in a separate step

controllers/keystoneservice_controller.go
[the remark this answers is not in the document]
abays: Ah, so it's actually two different finalizers that get added by the two different paths, because a different `helper` is being used in each case?

go.mod
```diff
@@ -36,10 +36,12 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
+	github.com/google/gnostic v0.5.7-v3refs // indirect
```
stuggi, line 39: this here seems to be the issue why client-go is required v0.24. need to check who pulls this in

go.mod
```diff
@@ -70,10 +72,10 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 	k8s.io/apiextensions-apiserver v0.23.0 // indirect
 	k8s.io/component-base v0.23.0 // indirect
-	k8s.io/klog/v2 v2.30.0 // indirect
-	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
-	k8s.io/utils v0.0.0-20211116205334-6203023598ed // indirect
-	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
+	k8s.io/klog/v2 v2.60.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42 // indirect
```
stuggi, line 76: this here is the root cause to pull in github.com/google/gnostic v0.5.7-v3refs

go.mod
stuggi: done, reverted the deps so that we can stick with 0.23.6

controllers/keystoneservice_controller.go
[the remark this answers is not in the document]
stuggi: yes that's correct

abays (approved):
/lgtm

PR #74, opened 2022-07-14
Api mod

dprince:


.github/workflows/build-keystone-operator.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
       uses: redhat-actions/openshift-tools-installer@v1
       with:
         source: github
-        operator-sdk: '1.18.1'
+        operator-sdk: '1.22.0'
```
stuggi, line 94: there is 1.22.1 , should go to that one, as we set in https://github.com/openstack-k8s-operators/install_yamls/pull/11 ?

.github/workflows/build-keystone-operator.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.17.x
+        go-version: 1.18.0
```
stuggi, line 85: As I mentioned in another PR from Andrew, I think in a follow up we should get this version from go.mod to not have to care about when things change there.

stuggi (approved):
/lgtm

PR #78, opened 2022-07-27
Make sure to cleanup service resources in keystone only with reference

stuggi:
When there is no ServiceID stored in the keystoneservice obj and
cleanup gets triggered e.g. other service endpoints get deleted
as well. This changes to only cleanup the resources in keystone
on obj delete when there is a ServiceID reference status.

Also minor updates to keystoneapi controller due to changes in lib-common to remove ReconcilerCommon


dprince (approved):

abays (approved):
/lgtm

PR #79, opened 2022-07-29
Adjust Keystone operator to lib-common reorg

abays:
Updates the operator to use the latest `lib-common` reorg layout.  Also fixes a bug with OpenStack service reconcile in `KeystoneService` controller where a "not found" (i.e. no existing service) during service creation was treated as an error.


stuggi (approved):
/lgtm

PR #82, opened 2022-08-05
Adds conditions the new way

stuggi:
* right now for the keystoneapi controller

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/39


stuggi:
Draft as keystoneservice is missing, but it shows on the keystoneapi how i is intended to work

api/v1beta1/keystoneapi_types.go
```diff
@@ -175,7 +175,7 @@ type KeystoneAPIStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[0].status",description="Status"
-//+kubebuilder:printcolumn:name="Reason",type="string",JSONPath=".status.conditions[0].reason",description="Reason"
```
abays, line 178: I see that the `KeystoneService` CRD still has `Status`, `Reason` and `Message` in its `printcolumns`.  Did you mean to remove `Reason` here from `KeystoneAPI` and not follow the same pattern?

controllers/keystoneapi_controller.go
```diff
@@ -113,7 +113,12 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	// initialize status
 	//
 	if instance.Status.Conditions == nil {
-		instance.Status.Conditions = condition.List{}
+		instance.Status.Conditions = condition.Conditions{}
+		instance.Status.Conditions.Init(nil)
+		// Register overall status immediately to have an early feedback e.g. in the cli
```
gibizer, line 118: This also helps avoiding that a later r.Update() call resets the instance.Status field. Probably we need to call r.Status().Update after every condition change to avoid that behavior

controllers/keystoneapi_controller.go
```diff
@@ -204,27 +214,26 @@ func (r *KeystoneAPIReconciler) reconcileInit(
...
+	ctrlResult, err := db.CreateOrPatchDB(
 		ctx,
 		helper,
 	)
-	instance.Status.Conditions.UpdateCurrentCondition(cond)
-
+	//instance.Status.Conditions.Set(condition.FalseCondition(condition.DBReadyCondition,db.GetDatabase().Status))
 	if err != nil {
+		//c := condition.FalseCondition(con)
+		//instance.Status.Conditions.Set(c)
```
gibizer, line 224: why it is comment out? it make sens to set a False conditions here

controllers/keystoneapi_controller.go
```diff
@@ -204,27 +214,26 @@ func (r *KeystoneAPIReconciler) reconcileInit(
...
 		return ctrl.Result{}, err
 	}
 	if (ctrlResult != ctrl.Result{}) {
-		r.Log.Info(cond.Message)
 		return ctrlResult, nil
```
gibizer, line 228: I guess this is a FalseCondition case too. Or the CreateOrPatchDB helper will set the conditions automatically?

controllers/keystoneapi_controller.go
```diff
@@ -204,27 +214,26 @@ func (r *KeystoneAPIReconciler) reconcileInit(
...
 	// wait for the DB to be setup
-	cond, ctrlResult, err = db.WaitForDBCreated(ctx, helper)
-	instance.Status.Conditions.UpdateCurrentCondition(cond)
+	ctrlResult, err = db.WaitForDBCreated(ctx, helper)
 	if err != nil {
 		return ctrlResult, err
 	}
 	if (ctrlResult != ctrl.Result{}) {
-		r.Log.Info(cond.Message)
 		return ctrlResult, nil
```
gibizer, line 237: ditto

controllers/keystoneapi_controller.go
abays: It does seem like this database logic should follow the same condition patterns as used for `DBSync`, `ExposeServices`, etc.

stuggi: This part is right now commented out as I wanted for this just to sync the condition on the mariadbdatabase object when that mariadb-operator controller reconciles and updates it. this would first required to get the new pattern into that operator. Until that is done I think I'll finis this part here to follow the pattern used for DBSync, ...

controllers/keystoneapi_controller.go
stuggi: yes, see above why I have not finished it, but will do now

controllers/keystoneapi_controller.go
stuggi: yes, see above why I have not finished it, but will do now

api/v1beta1/keystoneapi_types.go
stuggi: I have not finished the keystoneservice yet. wanted to submit this draft to get early feedback. will update it now

stuggi: changed to be in sync, I'll submit a new PR or at least a new commit for other changes to the keystoneservice controller

controllers/keystoneapi_controller.go
stuggi: done

controllers/keystoneapi_controller.go
stuggi: done

controllers/keystoneapi_controller.go
stuggi: done

controllers/keystoneapi_controller.go
gibizer: ack. if we set a condition then return, that will trigger the defer and persist the condition. So the only problematic scenario when we set a success condition, continue the current reconcile loop and call r.Update() before we return or before r.Status.Update() is called.

controllers/keystoneapi_controller.go
gibizer: ack. make sense

gibizer:
/lgtm
EnvTest would be nice to have for these changes to validate the state transitions and to prevent regressions in the future.

abays (approved):
/lgtm

stuggi:
> /lgtm EnvTest would be nice to have for these changes to validate the state transitions and to prevent regressions in the future.

Sure, I'll work on EnvTest for the keystone-operator as one of the next tasks.

PR #85, opened 2022-08-16
Add conditions to keystoneservice and initialize them

stuggi:
* Adds conditions to keystoneservice
* initializes conditions used by keystoneservice and keystoneapi at beginning to unknown status
* updates kesytoneapi deployment conditon to be true when `ReadyCount > 0`


api/v1beta1/conditions.go
```diff
@@ -0,0 +1,123 @@
...
+	KeystoneServiceOSUserReadyInitMessage = "Keystone Service user registration not started"
+
+	// KeystoneServiceOSUserReadyMessage
+	KeystoneServiceOSUserReadyMessage = "Keystone Service user %s ready"
+
+	// KeystoneServiceOSUserReadyWaitingMessage
+	KeystoneServiceOSUserReadyWaitingMessage = "Keystone Service user not yet ready"
+
+	// KeystoneServiceOSUserReadyErrorMessage
+	KeystoneServiceOSUserReadyErrorMessage = "Keystone Endpoints error occured %s"
```
abays, line 122: The value here looks like it should say something different.  Refer to service user instead of endpoints?

stuggi: right, good catch. will update

stuggi: done

abays (approved):
/lgtm

PR #90, opened 2022-09-13
Move external pkg to the api sub module

stuggi:
With moving the external pkg to the api sub module the dependency to github.com/openstack-k8s-operators/keystone-operator can be removed on other operators. They would then only depend on github.com/openstack-k8s-operators/keystone-operator/api and don't import anything used in github.com/openstack-k8s-operators/keystone-operator which is not required.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.0.0-20220816094529-135dc67c2cdf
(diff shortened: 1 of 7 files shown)


abays:
Makes sense to me.  Just curious though: what was the intention behind wrapping the `KeystoneService` in a wrapper helper?

stuggi:
> Makes sense to me. Just curious though: what was the intention behind wrapping the `KeystoneService` in a wrapper helper?

just to have all details to setup and manage the service in a struct type. The name was in the external package `KeystoneService` but that now clashes with the KeystoneService in the api itself.

PR #100, opened 2022-09-23
Split KeystoneService into Service and Endpoint

stuggi:
Some operators require to manage creating service, service user and assign the service user to the required role in a different component then register the service endpoints. Also there is the need to be able to create specific service endpoints via different components.
Therefore this change intruces a new CRD, KeystoneEndpoint, and moves the endpoint handling from the current KeystoneService CRD.

The KeystoneEndpoint controller gets the service ID to register the endpoints by requesting the corresponding KeystoneService object for the given ServiceName.

The new KeystoneEndpoint CRD allows, as it is today, to register multiple endpoints in keystone via the endpoint map in the spec, but allows to split into multiple KeystoneEndpoint CRs.


stuggi:
An example on how that is used can be seen in this commit https://github.com/openstack-k8s-operators/placement-operator/commit/2e01a44ad3251f0bf2b3d4652a16a8da62187118 . It splits creating the placement endpoints into a `placement-public` and `placement` KeystoneEndpoint.

api/v1beta1/keystoneendpoint.go
```diff
@@ -0,0 +1,200 @@
+/*
```
stuggi, line 1: this is just part of a split out of the func.go into keystoneapi.go, keystoneendpoint.go and keystoneservice.go

api/v1beta1/keystoneservice.go
```diff
@@ -21,107 +21,26 @@ import (
 	"time"
```
stuggi, line 21: this is just part of a split out of the func.go into keystoneapi.go, keystoneendpoint.go and keystoneservice.go

api/v1beta1/keystoneendpoint.go
```diff
@@ -0,0 +1,200 @@
...
+	namespace string,
+) (*KeystoneEndpoint, error) {
+
+	ke := &KeystoneEndpoint{}
+	err := h.GetClient().Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, ke)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			return ke, err
+		}
+		return ke, err
```
abays, line 168: There doesn't seem to be a difference between the `return` inside the `if` and the `return` outside of it

stuggi: right, not required. will remove the condition.

api/v1beta1/keystoneendpoint.go
```diff
@@ -0,0 +1,200 @@
...
+
+	ke, err := GetKeystoneEndpointWithName(ctx, h, name, namespace)
+
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
+
+	// Could get here if there was an k8s_errors.IsNotFound, so need to check again
+	// (we do it this way because we want to make sure the "RemoveFinalizer" call is
+	// executed even if the KeystoneEndpoint no longer exists)
```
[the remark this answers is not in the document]
stuggi: yes, at least the one inside the brackets.

api/v1beta1/keystoneendpoint.go
stuggi: done with commit `Simpify GetKeystone<*>WithName and DeleteKeystone<*>WithName`

api/v1beta1/keystoneendpoint.go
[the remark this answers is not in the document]
stuggi: done with commit `Simpify GetKeystone<*>WithName and DeleteKeystone<*>WithName`

controllers/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,422 @@
...
+				openstack.Endpoint{
+					Name:         instance.Spec.ServiceName,
+					ServiceID:    instance.Status.ServiceID,
+					Availability: availability,
+				},
+			)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+		}
```
abays, line 238: Is it possible that `instance.Spec.Endpoints` might currently have more endpoint keys than `instance.Status.EndpointIDs` such that then `os.DeleteEndpoint` might be called for an endpoint that doesn't actually exist on the backend?  Maybe it doesn't ultimately matter.  Just curious.

controllers/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,422 @@
...
+		return ctrl.Result{}, err
+	}
+
+	//
+	// Wait for KeystoneService is Ready and get the ServiceID from the object
+	//
+	ksSvc, err := keystonev1.GetKeystoneServiceWithName(ctx, helper, instance.Spec.ServiceName, instance.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
abays, line 275: I think this could hit `NotFound` errors here.  If so, would it be better to check for those and return a requeue result in such a situation?

stuggi: yes could return a specific requeue result to wait for some time and not an error. right now it returns an error and reconciles. will do

controllers/keystoneendpoint_controller.go
stuggi: `allEndpints` in https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/openstack/endpoint.go#L99 for that type should be empty and `os.DeleteEndpoint` won't call an actual delete. But I need to validate.

controllers/keystoneendpoint_controller.go
stuggi: added a commit with a change for that

api/v1beta1/keystoneendpoint.go
```diff
@@ -0,0 +1,200 @@
...
+	op, err := controllerutil.CreateOrPatch(ctx, h.GetClient(), endpoint, func() error {
+		endpoint.Spec = ke.endpoint.Spec
+		endpoint.Labels = util.MergeStringMaps(endpoint.Labels, ke.endpoint.Labels)
+
+		err := controllerutil.SetControllerReference(h.GetBeforeObject(), endpoint, h.GetScheme())
+		if err != nil {
+			return err
+		}
+
+		return nil
```
gibizer, line 90: Isn't it the same as simply `return err`

controllers/keystoneendpoint_controller.go
gibizer: If it is free to call delete on a not created endpoint then I think we can drop the `len(instance.Status.EndpointIDs) > 0` condition and the `else` branch too.

controllers/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,422 @@
...
+			)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	} else {
+		util.LogForObject(helper, fmt.Sprintf("Not deleting %s endpoints as there are no stored IDs", instance.Spec.ServiceName), instance)
+	}
+
+	// Service is deleted so remove the finalizer.
```
gibizer, line 243: Endpoints are deleted?

controllers/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,422 @@
...
+
+func (r *KeystoneEndpointReconciler) reconcileDelete(
+	ctx context.Context,
+	instance *keystonev1.KeystoneEndpoint,
+	helper *helper.Helper,
+	os *openstack.OpenStack,
+) (ctrl.Result, error) {
+	util.LogForObject(helper, "Reconciling Endpoint delete", instance)
+
+	// only cleanup the service if there are EndpointIDs references in the
```
gibizer, line 216: we are not deleting the service here.

controllers/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,422 @@
...
+		r.Scheme,
+		r.Log,
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Always patch the instance status when exiting this function so we can persist any changes.
+	defer func() {
+		// update the overall status condition if service is ready
```
gibizer, line 105: if endpoints are ready

controllers/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,422 @@
...
+
+func (r *KeystoneEndpointReconciler) reconcileNormal(
+	ctx context.Context,
+	instance *keystonev1.KeystoneEndpoint,
+	helper *helper.Helper,
+	os *openstack.OpenStack,
+) (ctrl.Result, error) {
+	util.LogForObject(helper, "Reconciling Endpoint normal", instance)
+
+	// If the service object doesn't have our finalizer, add it.
```
gibizer, line 262: // If the endpoint object

controllers/keystoneendpoint_controller.go
```diff
@@ -0,0 +1,422 @@
...
+
+func (r *KeystoneEndpointReconciler) reconcileEndpoints(
+	instance *keystonev1.KeystoneEndpoint,
+	helper *helper.Helper,
+	os *openstack.OpenStack,
+) error {
+	util.LogForObject(helper, "Reconciling Endpoints", instance)
+
+	// check for removed endpoints from the Spec.Endpoints definition
+	if instance.Status.EndpointIDs != nil && len(instance.Spec.Endpoints) < len(instance.Status.EndpointIDs) {
```
gibizer, line 331: If the client adds an endpoint and removes another in the same update call then the length of the EndpointIDs will be the same as the length of the Spec.Endpoint but the content will differ. I think in this case this code will not delete the removed endpoint.

stuggi: good call, let me improve that.

controllers/keystoneendpoint_controller.go
stuggi: done

controllers/keystoneendpoint_controller.go
stuggi: done

controllers/keystoneendpoint_controller.go
stuggi: done

controllers/keystoneendpoint_controller.go
stuggi: done

controllers/keystoneendpoint_controller.go
stuggi: yes, validated what I said and will add a commit to remove the conditionals as Gibi said above:

in Spec
~~~
# oc get keystoneendpoint placement -o json | jq .spec.endpoints
{
  "admin": "http://placement-admin-openstack.apps-crc.testing",
  "internal": "http://placement-internal-openstack.apps-crc.testing",
  "public": "http://placement-public-openstack.apps-crc.testing"
}
~~~

in keystone
~~~
# openstack endpoint list --service placement
+----------------------------------+-----------+--------------+--------------+---------+-----------+------------------------------------------------------+
| ID                               | Region    | Service Name | Service Type | Enabled | Interface | URL                                                  |
+----------------------------------+-----------+--------------+--------------+---------+-----------+------------------------------------------------------+
| 51d159e2b44e4fcd8c6a2a12348f8aa0 | regionOne | placement    | placement    | True    | admin     | http://placement-admin-openstack.apps-crc.testing    |
| 91fd568ccd1c43019a001c48c03904b0 | regionOne | placement    | placement    | True    | internal  | http://placement-internal-openstack.apps-crc.testing |
+----------------------------------+-----------+--------------+--------------+---------+-----------+------------------------------------------------------+
~~~

=> delete works fine

stuggi: done with 396862a08dea20cc594c19148a6bbb869bae79a5

abays:
+1 from me once all remaining comments from @gibizer are settled

controllers/keystoneendpoint_controller.go
stuggi: 5817c5c78e2ae547b4b0b7e3c962e1d542c9965b should cover this situation now

api/v1beta1/keystoneendpoint_types.go
```diff
@@ -0,0 +1,75 @@
...
+func init() {
+	SchemeBuilder.Register(&KeystoneEndpoint{}, &KeystoneEndpointList{})
+}
+
+// IsReady - returns true if endpoint got created ok in keystone
+// AND the endpoint ID registerd in the object status
+func (instance KeystoneEndpoint) IsReady() bool {
+
+	return instance.Status.Conditions.IsTrue(KeystoneServiceOSEndpointsReadyCondition) &&
+		len(instance.Status.EndpointIDs) == len(instance.Spec.Endpoints)
```
gibizer, line 74: I'm not sure we need other than the check for the ReadyCondition. But if we need to check on the Endpoints as there might be non reconciled changes between Spec and Status there, then a simple length equality is not enough. What if the URL changed within an endpoint in the Spec? The len() check does not detect that.

gibizer (commented):
Thanks for fixing my comments. I spotted one more place where we rely on the length of the Endpoints lists in IsReady().

api/v1beta1/keystoneendpoint_types.go
abays: I think right now that `status.EndpointIDs` stores a map of endpoint "type" (public, internal, admin) to an ID in Keystone, while the `spec.Endpoints` is a map of type to URL.  So if we wanted to compare those URLs to something in the status, perhaps we'll need to store the URLs in the status as well.  Basically right now can only compare endpoint types presence and absence.

stuggi: correct, as Andrew mentioned, right now its a map type -> endpoint ID. when an endpoint gets removed from the spec, the controller deletes the endpoint from keystone and when successful, also the ID gets deleted from the status [1]. So if we have the case with 2 endpoints, when one gets replaced and the delete of the endpoint is failing (for whatever reason), the spec number is 2 and the status.EndpointIDs also has count of 2. But the KeystoneServiceOSEndpointsReadyCondition will catch this. When an error happens during that delete, the KeystoneServiceOSEndpointsReadyCondition flips to false.

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/100/files#diff-33a1e743ae4c5cb0d0dad9e55e003bc15061dc9c169ad554a5368a956859102aR349

gibizer: If reconciliation error flips the KeystoneServiceOSEndpointsReadyCondition then that also support my point that the check for length is not needed here. 

It all depends on how precise we want to be with the IsReady().

- a) We want to say IsReady() -> False if the Spec.Endpoints was updated but the reconcile loop hasn't been triggered yet. Then we need to compare Spec.Endpoints with Status.Endpoints here to be able to detect if there was a change that needs reconciliation. But if we want to do this then simply checking the length is not enough as might be the URL changed in an existing Endpoint. I understand that this is not possible today as the Status.Endpoint does not have the URL.

- b) We are OK to keep IsReady() -> True if the Spec.Endpoints was updated but the reconcile loop hasn't been triggered yet. In this case the 
```
IsReady() {return instance.Status.Conditions.IsTrue(KeystoneServiceOSEndpointsReadyCondition)}
```
is enough.

stuggi: 32e45f2bfe61883678747b03c2257f67f93f15da

gibizer (commented):
/lgtm
thanks!

PR #101, opened 2022-09-23
Move rendered CRD files dir to /api/crd and link

stuggi:
With the new api sub module, consumer operators just have to import the api submodule. For envtest it is required to install CRD for e.g. keystone-operator and mariadb-operator to be able to use it.

There are multiple ways this could be done.
1) like in this patch move the directory config/crd/bases which
   holds the generated files in to the api module that they are
   available with the module
2) load the base module in the envtest setup, but this requires
   that the base and sub module matches with each released version
3) always add the required CRD files into the operator which needs
   it.
4) generate the CRD files in the envtest setup from the api submodule
   it has using controller-gen.
5) update the Makefile to generate the files in the new location.
   Need to be checked if that would break any other sdk functionality
6) Do a complete other approach, like other projects do, and move
   all api modules into their own repo [1]. Need to be checked what
   was the root cause for this and if that is something which would
   help us.

To unblock e.g. the placement-operator this patch does 1) until we have a long term solution, which might still be 1).

[1] https://github.com/openshift/api


gibizer:
looks good to me. Thanks @stuggi !

gibizer (commented):
looks good to me. Thanks @stuggi !

stuggi:
rebased after https://github.com/openstack-k8s-operators/keystone-operator/pull/100 merged

gibizer:
Still looks good after the rebase.

abays (approved):
/lgtm

PR #112, opened 2022-10-15
Cleanup jobs after hash is persisted

gibizer:
There was a lib-common issue where job.DoJob() deleted the Job right after it is succeeded. Then the controller tried to store the hash of the Job in Status. But if that failed (due to e.g. Conflict) then in the next Reconcile call lib-common would re-create and re-run the Job as the hash of the previous Job and the Job itself is lost.

The lib-common is fixed in a way that the automatic Job deletion is removed from DoJob() and that needs to be now done on the caller side via the new job.DeleteAllSuccededJobs() call.

This patch adapts to the new lib-common version by doing the Job cleanup right after the Status.Update() successfully persisted the Job's hash.

Depends-On: openstack-k8s-operators/lib-common#77

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -280,8 +280,7 @@ func (r *KeystoneAPIReconciler) reconcileInit(
 	dbSyncjob := job.NewJob(
 		jobDef,
 		keystonev1.DbSyncHash,
-		instance.Spec.PreserveJobs,
-		5,
+		time.Duration(5)*time.Second,
 		dbSyncHash,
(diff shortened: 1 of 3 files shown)


gibizer:
So after this fix was still able to reproduce occasional double Job execution. It turned out that we can read stale data from a cache in operator-sdk. So simply moving the deletion logic after the Job hash is persisted is not enough. @abays suggested to simply delay the Job deletion by setting ttlSecondsAfterFinished on the Job. So with a big enough ttl value we can circumvent this issue.

PR #113, opened 2022-10-19
Only add finalizer and update instance if necessary

abays:
We should only attempt to add finalizers in reconcileNormal() if they are missing from the instance being reconciled, and if such action is needed, we should end that reconcile after updating the instance to avoid potentially reconciling again with an outdated version of the instance (see [1] for details).

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/1#discussion_r997014701


stuggi (approved):
/lgtm

PR #115, opened 2022-11-02
Add e2e job for keystone operator

Sandeepyadav93:
In Pr[1], we added a reusable workflow that build, install and test operator. With this commit, we are adding e2e job for keystone-operator by consuming the reusable workflow.

Keeping security in mind, By default, we will only trigger CI job if someone is collaborator/have write access on repo otherwise they will need a `ok-to-test` label from reviewer to run e2e job.

[1] https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/53

### DIFF
--- a/.github/workflows/label-on-pr.yaml
+++ b/.github/workflows/label-on-pr.yaml
@@ -1,0 +1,15 @@
+name: Workflow on label addition
+
+on:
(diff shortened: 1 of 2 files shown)


stuggi (approved):
/lgtm

PR #132, opened 2022-12-01
Return immediately after status initialization

abays:
Whenever we call Update for a CR, we should end the current reconciliation and return immediately to avoid potentially losing the change in the next reconcile loop due to a race condition described in [1]. This pattern should be repeated across all our operators.

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/1#discussion_r997014701


stuggi (commented):
how about the other status.update() calls. should we also return there?

abays:
> how about the other status.update() calls. should we also return there?

I was planning to address those later and only deal with the status initialization here, but I suppose it makes sense to just go through the operator now and address this concern wherever it appears.  I will switch this to draft and work on that.

controllers/keystoneapi_controller.go
```diff
@@ -109,6 +109,26 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		return ctrl.Result{}, err
+	}
+
+	if instance.DeletionTimestamp.IsZero() && !controllerutil.ContainsFinalizer(instance, helper.GetFinalizer()) {
+		// If the service object doesn't have our finalizer, add it.
+		controllerutil.AddFinalizer(instance, helper.GetFinalizer())
+		// Register the finalizer immediately to avoid orphaning resources on delete
+		err := r.Update(ctx, instance)
+
+		return ctrl.Result{}, err
```
stuggi, line 129: instead of use `controllerutil.ContainsFinalizer` could do this in one call and check the return of `controllerutil.AddFinalizer`:
~~~
It returns an indication of whether it updated the object's list of finalizers.
~~~

```
// Register the finalizer immediately to avoid orphaning resources on delete
if controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
  err := r.Update(ctx, instance)
  return ctrl.Result{}, err
}
```

abays: Ah, I had never read the function description.  That sounds good and I will change that.

controllers/keystoneapi_controller.go
```diff
@@ -317,10 +323,8 @@ func (r *KeystoneAPIReconciler) reconcileInit(
 	}
 	if dbSyncjob.HasChanged() {
 		instance.Status.Hash[keystonev1.DbSyncHash] = dbSyncjob.GetHash()
-		if err := r.Client.Status().Update(ctx, instance); err != nil {
-			return ctrl.Result{}, err
-		}
 		r.Log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.Hash[keystonev1.DbSyncHash]))
+		return ctrl.Result{}, nil
```
stuggi, line 327: With this we'll only reconcile on the next event, right? for running jobs we can expect one to come. just wondering if there can be a situation in one of the bellow `return ctrl.Result{}, nil` that this might not happen? It might make sense to set `Requeue` to true ? [1]

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/reconcile#Result

controllers/keystoneapi_controller.go
```diff
@@ -526,9 +519,12 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
 	// create hash over all the different input resources to identify if any those changed
 	// and a restart/recreate is required.
 	//
-	inputHash, err := r.createHashOfInputHashes(ctx, instance, configMapVars)
+	inputHash, hashChanged, err := r.createHashOfInputHashes(ctx, instance, configMapVars)
 	if err != nil {
 		return ctrl.Result{}, err
+	} else if hashChanged {
+		// Hash changed and instance status should be updated, so we need to return and reconcile again
+		return ctrl.Result{}, nil
```
stuggi, line 527: With this we'll only reconcile on the next event, right? for running jobs we can expect one to come. just wondering if there can be a situation in one of the bellow `return ctrl.Result{}, nil` that this might not happen? It might make sense to set `Requeue` to true ? [1]

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/reconcile#Result

abays: Yeah, good point.  We should force the requeue whenever we update the status and need it to save and then continue reconciling again.

controllers/keystoneapi_controller.go
```diff
@@ -109,6 +109,25 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// If we're not deleting this and the service object doesn't have our finalizer, add it.
+	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
+		// Register the finalizer immediately to avoid orphaning resources on delete
+		err := r.Update(ctx, instance)
+
+		return ctrl.Result{}, err
```
stuggi, line 128: shouldn't we requeue here as well?

controllers/keystoneapi_controller.go
```diff
@@ -127,9 +146,7 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		instance.Status.Conditions.Init(&cl)
 
 		// Register overall status immediately to have an early feedback e.g. in the cli
-		if err := r.Status().Update(ctx, instance); err != nil {
-			return ctrl.Result{}, err
-		}
+		return ctrl.Result{}, r.Status().Update(ctx, instance)
```
stuggi, line 149: same, shouldn't we requeue here as well?

controllers/keystoneapi_controller.go
abays: I think in these cases that we don't need to explicitly include `Requeue: true` because the `Update()` of the instance will already have caused the instance to be requeued

bogdando: when we update instance with added finalizer to be set, the next reconcile from a queue could still not see it commited by kube-apisserver, right? So we would add it here once again, call update again, and the story repeats with a busy loop again and again. To prevent that, we probably should not add a finalizer unconditionally *but which condition to use, if all we have is cached view into the system state*?

bogdando: Is there a way to get information about the context this reconcile has been called with? Like requed by us earlier, or auto-requed after Update() or Patch() call, or triggered by watcher? Having that info might help to aboid the busy loop I described above.

But it shouldn't happen, if there is guarantees that informer/watcher do not place a reconcile event for added finalizer until kube-apiserver applies it and it pops up updated in the cache.

bogdando: @mdbooth WDYT ^^?

bogdando: > As soon as you call Update() here the watch is going to update the shared informer, and this reconciler is queued to be called again with the state of the object at this point.

So if I get that right, the queued reconcile for this Update() call that adds a finalizer, is guaranteed to have that finalizer in the cache by the time we hit this code again.

still not sure if we should be adding finalizer unconditionally, as there could be later reconcile loops unrelated to that Update() for finalizer? And then, we should prefer to not add it again and avoid generating unnecessary updates in the queue?..

abays: Just to respond to one point specifically, adding the finalizer as we are doing here is no unconditional.  The docs [1] state:

```
AddFinalizer accepts an Object and adds the provided finalizer if not present. It returns an indication of whether it updated the object's list of finalizers. 
```

The second sentence leads me to believe that `AddFinalizer` will only return true if the instance indeed lacked the finalizer and thus needed it to be added.

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#AddFinalizer

gibizer, line 128: If we have Update + return, then we can simply have just a return here and let the deferred call do the Update. This would limit the places we do Updates on ourselves. See nova-controller as an example:
- https://github.com/openstack-k8s-operators/nova-operator/blob/fc2fe084dc1031a0d6582a14f59562f88fc70305/controllers/nova_controller.go#L109-L149
- https://github.com/openstack-k8s-operators/nova-operator/blob/fc2fe084dc1031a0d6582a14f59562f88fc70305/controllers/nova_controller.go#L159-L167

gibizer: I agree that update on ourselves will trigger a new reconcile so no requeue is needed.

controllers/keystoneapi_controller.go
gibizer: I think update on ourselves will trigger a reconcile anyhow.
Also instead of doing an update here, I would just return and let the defer call do the update in a single place

controllers/keystoneapi_controller.go
gibizer: don't we get reconciled because we changed our own Status?

controllers/keystoneapi_controller.go
```diff
@@ -819,22 +814,23 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
 	envVars map[string]env.Setter,
-) (string, error) {
+) (string, bool, error) {
+	var hashMap map[string]string
+	changed := false
 	mergedMapVars := env.MergeEnvs([]corev1.EnvVar{}, envVars)
 	hash, err := util.ObjectHash(mergedMapVars)
 	if err != nil {
-		return hash, err
+		return hash, changed, err
```
bogdando, line 829: shouldn't it retutn nil instead of changed=false? an error means no changed, not unchanged state

controllers/keystoneapi_controller.go
bogdando: Is relying on hashChanged evaluation safe here? I see it takes the instance as an argument, so wouldn't the mentioned earlier read-after-write (stale reads) race condition happening here?

controllers/keystoneapi_controller.go
abays: We could change the bool to `*bool` and then be able to return `nil`, but as it at the moment, we have to return either true or false.

controllers/keystoneapi_controller.go
abays: @gibizer Yes, my Github comment is outdated and the code comment is what is proper here

bogdando: thank you for explaining

controllers/keystoneapi_controller.go
abays: Yes, we should let the `defer()` update do the actual call as much as possible.  I will try that out here.

controllers/keystoneapi_controller.go
abays: Going to try that

controllers/keystoneapi_controller.go
```diff
@@ -526,9 +531,13 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
 	// create hash over all the different input resources to identify if any those changed
 	// and a restart/recreate is required.
 	//
-	inputHash, err := r.createHashOfInputHashes(ctx, instance, configMapVars)
+	inputHash, hashChanged, err := r.createHashOfInputHashes(ctx, instance, configMapVars)
 	if err != nil {
 		return ctrl.Result{}, err
+	} else if hashChanged {
+		// Hash changed and instance status should be updated (which will be done by main defer func),
+		// so we need to return and reconcile again
```
bogdando, line 539: please add a log msg as the comment explains

controllers/keystoneapi_controller.go
```diff
@@ -411,9 +428,6 @@ func (r *KeystoneAPIReconciler) reconcileInit(
 	}
 	if bootstrapjob.HasChanged() {
 		instance.Status.Hash[keystonev1.BootstrapHash] = bootstrapjob.GetHash()
-		if err := r.Client.Status().Update(ctx, instance); err != nil {
-			return ctrl.Result{}, err
-		}
 		r.Log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.Hash[keystonev1.BootstrapHash]))
```
bogdando, line 431: as you removed an immediate update, this message becomes no longer relevant - nothing has been added by far?

controllers/keystoneapi_controller.go
```diff
@@ -819,22 +828,23 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
-		return hash, err
+		return hash, changed, err
 	}
-	if hashMap, changed := util.SetHash(instance.Status.Hash, common.InputHashName, hash); changed {
+	if hashMap, changed = util.SetHash(instance.Status.Hash, common.InputHashName, hash); changed {
 		instance.Status.Hash = hashMap
-		if err := r.Client.Status().Update(ctx, instance); err != nil {
-			return hash, err
-		}
 		r.Log.Info(fmt.Sprintf("Input maps hash %s - %s", common.InputHashName, hash))
```
bogdando, line 847: ditto, w/o updaiting it right away, this log message is too early. It could be rephrased like queued update, or something

controllers/keystoneapi_controller.go
```diff
@@ -158,17 +130,68 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-		if changed := helper.GetChanges()["status"]; changed {
-			patch := client.MergeFrom(helper.GetBeforeObject())
+		changes := helper.GetChanges()
+		patch := client.MergeFrom(helper.GetBeforeObject())
+
+		if changes["metadata"] {
+			err := r.Client.Patch(ctx, instance, patch)
+			if k8s_errors.IsConflict(err) {
+				util.LogForObject(helper, "Metadata update conflict", instance)
+				return
```
gibizer, line 143: I noted this over the nova-operator PR but noting it here too. So I think if we need to signal failure from the defered call to the Reconcile call, then we need to change the return value of the Reconcile call from here instead of just returning. I think the reconcile logic uses the returned err from Reconcile to a) log b) requeue the reconciliation.

To change the return value of Reconcile from here we have to name the return value of Reconcile in its signature and then use that variable name here in the deferred call to change the actual return value of Reconcile.

controllers/keystoneapi_controller.go
```diff
@@ -158,17 +130,68 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			} else if err != nil && !k8s_errors.IsNotFound(err) {
+				util.LogErrorForObject(helper, err, "Metadate update failed", instance)
+				return
```
gibizer, line 146: ditto

controllers/keystoneapi_controller.go
```diff
@@ -158,17 +130,68 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+		}
 
-			if err := r.Status().Patch(ctx, instance, patch); err != nil && !k8s_errors.IsNotFound(err) {
-				util.LogErrorForObject(helper, err, "Update status", instance)
+		if changes["status"] {
+			err := r.Client.Status().Patch(ctx, instance, patch)
+			if k8s_errors.IsConflict(err) {
+				util.LogForObject(helper, "Status update conflict", instance)
+				return
```
gibizer, line 154: ditto

controllers/keystoneapi_controller.go
```diff
@@ -158,17 +130,68 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+			} else if err != nil && !k8s_errors.IsNotFound(err) {
+				util.LogErrorForObject(helper, err, "Status update failed", instance)
+				return
```
gibizer, line 158: ditto

gibizer (changes requested):

controllers/keystoneapi_controller.go
abays: Well, in this case the message is merely displaying the hash as opposed to including a verb like `added` to indicate that something might have been persisted.  But we could qualify the message with "(queued for update)".

I'm thinking we need some sort of the logging in the upcoming standardized `defer`-ed instance update func ([example](https://github.com/openstack-k8s-operators/lib-common/compare/master...abays:lib-common:helper_instance_update)) to indicate that the instance (and/or its status) was successfully patched/updated at the end of the reconcile loop.

controllers/keystoneapi_controller.go
abays: Got it.  Thank you for noting and explaining this.

abays: Done

controllers/keystoneapi_controller.go
abays: Done

controllers/keystoneapi_controller.go
abays: Done

controllers/keystoneapi_controller.go
abays: Done

stuggi (commented):
looks good to me

gibizer (commented):
looks good. Thanks @abays !

gibizer (approved):

PR #135, opened 2022-12-07
Ensure services and endpoints are removed before API

abays:
Ensure that `KeystoneEndpoint` and `KeystoneService` resources are deleted before the `KeystoneAPI` resource is removed


stuggi (commented):
looks good to me

controllers/keystoneendpoint_controller.go
```diff
@@ -235,6 +250,15 @@ func (r *KeystoneEndpointReconciler) reconcileDelete(
 		}
 	}
 
+	// Remove the finalizer for this endpoint from the KeystoneAPI
+	if controllerutil.RemoveFinalizer(keystoneAPI, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
+		err := r.Update(ctx, keystoneAPI)
+
+		if err != nil {
+			return ctrl.Result{}, err
```
bogdando, line 258: ditto

bogdando (changes requested):

controllers/keystoneendpoint_controller.go
abays: Ah, I actually misread my own changes when responding to this concern earlier.  The objected being updated here is not the instance being reconciled (we're updating `KeystoneAPI`, not the `KeystoneEndpoint` currently being processed), so the race condition paradigm does not apply.

bogdando (approved):

PR #146, opened 2023-01-05
Add some KUTTL tests for keystone operator

cescgina:
These tests cover basic operations: deployment, scaling up, scaling down and clean up. These tests use code from the install_yamls repo to reuse some operation like the deployment of the keystone and mariadb db operators, as well as the cleanup from previous runs.

Tests are thought to be run from the `install_yaml` repo, see the following PR: https://github.com/openstack-k8s-operators/install_yamls/pull/50


tests/kuttl/tests/keystone_scale/03-assert.yaml
```diff
@@ -0,0 +1,63 @@
...
+# We can't provide 3 separate Pod YAMLs to match against because we don't know
+# what random strings were used in the Pod names for the KeystoneAPI, so instead
+# we run a bash script to check for the 3 Pods
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: |
+      set -x
+      sleep 3
```
gibizer, line 57: Why we need a sleep here? As far as I know kuttl waits for 30sec to reach the expected end state.

tests/kuttl/tests/keystone_scale/04-scale-down-keystoneapi.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      oc patch keystoneapi -n openstack keystone --type='json' -p='[{"op": "replace", "path": "/spec/replicas", "value":1}]'
```
gibizer, line 5: Could you add a separate step where you assert that scaling down to 0 replicas works as well?

tests/kuttl/tests/keystone_scale/04-assert.yaml
```diff
@@ -0,0 +1,63 @@
...
+  readyCount: 1
+---
+# We can't provide 3 separate Pod YAMLs to match against because we don't know
+# what random strings were used in the Pod names for the KeystoneAPI, so instead
+# we run a bash script to check for the 1 Pods
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: |
```
gibizer, line 55: This is now repeated from the previous assert. Can we pull this out to a common thing?

tests/kuttl/tests/keystone_scale/03-assert.yaml
```diff
@@ -0,0 +1,63 @@
...
+      podcount=$(oc get pods -n openstack -l service=keystone -o name | wc -l)
+      if [ "$podcount" == "3" ]; then
```
gibizer, line 59: This fails to me with:

```
logger.go:42: 15:23:26 | keystone_scale/3-scale-keystoneapi | keystoneapi.keystone.openstack.org/keystone patched
I0105 15:23:27.246030  427800 request.go:655] Throttling request took 1.049253403s, request: GET:https://api.crc.testing:6443/apis/flowcontrol.apiserver.k8s.io/v1beta2?timeout=32s
    logger.go:42: 15:23:29 | keystone_scale/3-scale-keystoneapi | running command: [sh -c set -x
        sleep 3
        podcount=$(oc get pods -n openstack -l service=keystone -o name | wc -l)
        if [ "$podcount" == "3" ]; then
          exit 0
        else
          exit 1
        fi
        ]
    logger.go:42: 15:23:29 | keystone_scale/3-scale-keystoneapi | + sleep 3
    logger.go:42: 15:23:32 | keystone_scale/3-scale-keystoneapi | + oc get pods -n openstack -l service=keystone -o name
    logger.go:42: 15:23:32 | keystone_scale/3-scale-keystoneapi | + wc -l
    logger.go:42: 15:23:32 | keystone_scale/3-scale-keystoneapi | + podcount=3
    logger.go:42: 15:23:32 | keystone_scale/3-scale-keystoneapi | + [ 3 == 3 ]
    logger.go:42: 15:23:32 | keystone_scale/3-scale-keystoneapi | sh: 4: [: 3: unexpected operator
    logger.go:42: 15:23:32 | keystone_scale/3-scale-keystoneapi | + exit 1
```

tests/kuttl/tests/keystone_scale/03-assert.yaml
```diff
@@ -0,0 +1,63 @@
...
+  readyCount: 3
+---
+# We can't provide 3 separate Pod YAMLs to match against because we don't know
+# what random strings were used in the Pod names for the KeystoneAPI, so instead
+# we run a bash script to check for the 3 Pods
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: |
```
gibizer, line 55: Can we somehow avoid scripting by doing a partial match via https://kuttl.dev/docs/testing/asserts-errors.html#listing-resources-in-the-cluster ?

gibizer (changes requested):
I've limited kuttl knowledge so take my inline comments with a grain of salt. Overall I think it is a great start of having integration test coverage. However I have some questions / suggestions inline.

tests/kuttl/tests/keystone_scale/01-assert.yaml
```diff
@@ -0,0 +1,62 @@
...
+  containers:
+  - image: quay.io/tripleozedcentos9/openstack-mariadb:current-tripleo
+    imagePullPolicy: IfNotPresent
+    name: mariadb
+    resources: {}
+  restartPolicy: Always
+  serviceAccount: mariadb-operator-mariadb
+  serviceAccountName: mariadb-operator-mariadb
+status:
+  phase: Running
```
stuggi, line 62: if possible this should be in the mariadb-operator repo and can be validated by some `make mariadb_deploy_validate` or something. like with `make mariadb_deploy` it could then be consumed by any other service operator and if there is a change we won't have to update all asserts in all the operators.

We should have this for all the operators for the basic setup make <service>_deploy. with this installing
the dependencies for an operator and validate that the service is ok would be easy to implement and maintained.

tests/kuttl/tests/keystone_scale/02-assert.yaml
```diff
@@ -0,0 +1,134 @@
...
+      value: keystone
+    image: quay.io/tripleozedcentos9/openstack-keystone:current-tripleo
+    imagePullPolicy: IfNotPresent
+    name: init
+    resources: {}
+  restartPolicy: Always
+  serviceAccount: keystone-operator-keystone
+  serviceAccountName: keystone-operator-keystone
+status:
+  phase: Running
```
stuggi, line 134: like above, almost all other service operators need keystone. and they'd call `make keystone_deploy` to get it up. I think we should have a corresponding `make keystone_deploy_validate` which uses this assert file to check. wondering if it is best to keep this in a common place/name and just link it here to 02-assert.yaml and then could use the common one in `make keystone_deploy_validate`?

tests/kuttl/tests/keystone_scale/01-assert.yaml
stuggi: or have a make target which copies this base validation assert file from the mariadb-operator here where it is needed to be checked...

tests/kuttl/tests/keystone_scale/02-assert.yaml
stuggi: or as above copy it then to the consuming operator.

tests/kuttl/tests/keystone_scale/03-assert.yaml
```diff
@@ -0,0 +1,63 @@
...
+  apiEndpoint:
+    admin: http://keystone-admin-openstack.apps-crc.testing
+    internal: http://keystone-internal-openstack.apps-crc.testing
+    public: http://keystone-public-openstack.apps-crc.testing
+  databaseHostname: openstack
+  readyCount: 3
+---
+# We can't provide 3 separate Pod YAMLs to match against because we don't know
+# what random strings were used in the Pod names for the KeystoneAPI, so instead
+# we run a bash script to check for the 3 Pods
```
stuggi, line 50: if it is just the pod count we check , we maybe just check the deployment?

tests/kuttl/tests/keystone_scale/04-assert.yaml
```diff
@@ -0,0 +1,63 @@
...
+  secret: osp-secret
+status:
+  apiEndpoint:
+    admin: http://keystone-admin-openstack.apps-crc.testing
+    internal: http://keystone-internal-openstack.apps-crc.testing
+    public: http://keystone-public-openstack.apps-crc.testing
+  databaseHostname: openstack
+  readyCount: 1
+---
+# We can't provide 3 separate Pod YAMLs to match against because we don't know
```
stuggi, line 48: same could check the replicas/readyReplicas of the deployment?

tests/kuttl/tests/keystone_scale/01-assert.yaml
abays: I agree.  For common dependency operators like MariaDB and Keystone, we need to have some way of supporting a centralized definition of "success" for their deployment.  i.e. when I was the `install_yamls` KUTTL example, I did something like this: https://github.com/abays/install_yamls/blob/kuttl/tests/kuttl/common/tests/assert/deploy-keystone.yaml.  The question in this distributed context becomes how/where do we store/represent these common objects?

tests/kuttl/tests/keystone_scale/04-assert.yaml
abays: Yes, we could check the deployment if we want to use that as the source of truth for the state of the pods.  I would assume the states are always consistent between the two given that it's such a core k8s paradigm.

tests/kuttl/tests/keystone_scale/03-assert.yaml
cescgina: You're right, it's not needed, I've removed them

kuttl-test.yaml
```diff
@@ -0,0 +1,25 @@
...
+# EXECUTION (from install_yamls repo root):
+#
+#   make kuttl_keystone
+#
+# ASSUMPTIONS:
+#
+# 1. Latest version of kuttl is installed at /usr/local/bin/kubectl-kuttl
+#    - wget https://github.com/kudobuilder/kuttl/releases/download/v0.11.1/kubectl-kuttl_0.11.1_linux_x86_64
+#    - mv kubectl-kuttl_0.11.1_linux_x86_64 /usr/local/bin/kubectl-kuttl
+#    - chmod 755 /usr/local/bin/kubectl-kuttl
```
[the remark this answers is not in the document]
gibizer: Thanks for checking. We can keep it as is.

tests/kuttl/tests/keystone_scale/04-scale-down-keystoneapi.yaml
cescgina: Done!

tests/kuttl/tests/keystone_scale/03-assert.yaml
cescgina: Following the suggestion from @stuggi, I've changed this code to check the number of pods from the deployment, so no more bash here and the error should be gone.

tests/kuttl/tests/keystone_scale/03-assert.yaml
cescgina: Done, now the pod count is checked from the deployment.

tests/kuttl/tests/keystone_scale/04-assert.yaml
cescgina: Done

tests/kuttl/tests/keystone_scale/04-assert.yaml
cescgina: After last change, this small script is no longer used

tests/kuttl/tests/keystone_scale/04-scale-down-keystoneapi.yaml
gibizer: Awesome. Thanks!

tests/kuttl/tests/keystone_scale/03-assert.yaml
```diff
@@ -0,0 +1,53 @@
...
+    public: http://keystone-public-openstack.apps-crc.testing
+  databaseHostname: openstack
+  readyCount: 3
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: keystone
+spec:
+  replicas: 3
```
gibizer, line 53: This only asserts that keystone operator changed the replica count on the deployment to 3. But it does not wait to see if 3 replicas are actually possible to start up successfully. To assert that:
```yaml
status:
    availableReplicas: 3
```
will wait for the 3 Pods to be ready.

gibizer (changes requested):
Nice. I was able to replicate the test execution locally without any problem. I can even re-run it multiple times. I left some suggestion inline.

tests/kuttl/tests/keystone_scale/03-assert.yaml
cescgina: That makes a lot of sense, added. Thanks for the suggestion!

tests/kuttl/tests/keystone_scale/05-assert.yaml
```diff
@@ -0,0 +1,52 @@
...
+    internal: http://keystone-internal-openstack.apps-crc.testing
+    public: http://keystone-public-openstack.apps-crc.testing
+  databaseHostname: openstack
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: keystone
+spec:
+  replicas: 0
```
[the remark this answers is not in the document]
cescgina: I've tried adding it, but then the tests fail because the `availableReplicas` key is not found. I've done `oc get deployment keystone -o yaml` and the key is not there when the deployment is scaled down to 0 replicas, but it appears if the number of replicas are 1 or more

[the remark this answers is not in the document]
gibizer: yeah that is probably the infamous omitempty behavior of the k8s. The '0' value is the default / empty value of an integer field. And if a field is defined with the omitempty tag then k8s will not return the field if the value of the field is the empty value of the type of the field.

I'm not sure kuttl can taught that a missing field here is equal to the 0 value.  I opened an issue on kuttl to see if the devs there now a way to solve this. https://github.com/kudobuilder/kuttl/issues/455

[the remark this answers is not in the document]
gibizer: Based on the reply here https://github.com/kudobuilder/kuttl/issues/455#issuecomment-1377204077 we can sort of assert it by saying it is an error if there is such a field provided. I haven't tested this idea but it might work.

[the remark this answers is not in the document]
cescgina: I'm testing right now. It seems that we can use both an assert in the same test step. But the example errors does not seem to work. It seems to be looking for the string `{}` and not any value. So if I patch keystone to scale to 1, then it does not fail. I don't think kuttl allows partial/regex matching to check for a value that is not 0.
One solution would be to just check for a value of `availableReplicas` of 1, which is  the value from the previous test step. It seems brittle, but it might do the job, other than that, we would probably need some script to detect it.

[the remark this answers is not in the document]
gibizer: @cescgina thanks for trying out. Could you please bring back this to the original thread https://github.com/kudobuilder/kuttl/issues/455#issuecomment-1377204077 maybe the devs there has further ideas.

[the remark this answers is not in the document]
cescgina: Apparently there is no way to do it just with kuttl, I've pushed a commit with some scripting to do the check.

cescgina:
@stuggi @abays I've tried to refactor the tests and the Makefile from `install_yamls` to make it easier to reuse the asserts and validate the deployments from keystone and mariadb operators. The mariadb tests are now in the mariadb-operator repo (PR: https://github.com/openstack-k8s-operators/mariadb-operator/pull/48).

tests/kuttl/common/assert_keystone_deployment.yaml
```diff
@@ -0,0 +1,134 @@
...
+  readyCount: 1
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  annotations:
+    openshift.io/scc: anyuid
+  labels:
+    service: keystone
+  #name: keystone-69f88774bd-gjspj   Can't include name due to random append
```
stuggi, line 55: the operators created the deployment, not the pod. should we instead check if the deployment and its pod spec in there is correct and the readycount for the pods?

tests/kuttl/common/assert_keystone_deployment.yaml
```diff
@@ -0,0 +1,134 @@
+#
+# Check for:
+#
+# - 1 KeystoneAPI CR
+# - 1 Pod for KeystoneAPI CR
```
stuggi, line 5: might also want to check for the created service and route?

tests/kuttl/tests/keystone_scale/02-assert.yaml
```diff
@@ -0,0 +1,55 @@
...
+  databaseUser: keystone
+  debug:
+    bootstrap: false
+    dbSync: false
+    service: false
+  passwordSelectors:
+    admin: AdminPassword
+    database: KeystoneDatabasePassword
+  preserveJobs: false
+  region: regionOne
```
stuggi, line 33: could remove everything which hasn't changed, like for the deployment bellow. but could also leave it.

tests/kuttl/tests/keystone_scale/03-assert.yaml
```diff
@@ -0,0 +1,55 @@
...
+  databaseUser: keystone
+  debug:
+    bootstrap: false
+    dbSync: false
+    service: false
+  passwordSelectors:
+    admin: AdminPassword
+    database: KeystoneDatabasePassword
+  preserveJobs: false
+  region: regionOne
```
stuggi, line 33: same as above, could remove parts which has not changed and keep the ones we explicite want to check.

tests/kuttl/tests/keystone_scale/04-assert.yaml
```diff
@@ -0,0 +1,65 @@
...
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: keystone
+spec:
+  replicas: 0
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
```
stuggi, line 55: I think instead of running this TestAssert, we could also use errors and check if there are pods with labels for this CRD. 
~~~
  labels:
    service: keystone
~~~

tests/kuttl/tests/keystone_scale/05-errors.yaml
```diff
@@ -0,0 +1,55 @@
+#
+# Check for:
+#
+# No KeystoneAPI CR
+# No Pods for KeystoneAPI CR
```
stuggi, line 5: also check for KeystoneAPI Route and Service gone

tests/kuttl/tests/keystone_scale/05-errors.yaml
```diff
@@ -0,0 +1,55 @@
...
+      cpu: "1"
+      memory: 500Mi
+  secret: osp-secret
+status:
+  apiEndpoint:
+    admin: http://keystone-admin-openstack.apps-crc.testing
+    internal: http://keystone-internal-openstack.apps-crc.testing
+    public: http://keystone-public-openstack.apps-crc.testing
+  databaseHostname: openstack
+  readyCount: 3
```
stuggi, line 45: just metadata should be enough

stuggi (commented):
looks good, just some ideas/question on things we could improve to make it easier to maintain, as we have seen from OSPdO. A thing we have seen there and I think we might want to do different is, if things in the CRD change its a mess to update all assert files in all tests. So we might want to keep only the important parts we want to check in the steps. For this PR I think we don't have to keep the full keystoneapi spec in all asserts and just keep the one for scaling, as this is what the test is for, right?
if there are other functional tests, to validate if the operator behaves correct, this will be a different test and have the spec asserts there.

tests/kuttl/common/assert_keystone_deployment.yaml
```diff
@@ -0,0 +1,134 @@
+#
```
stuggi, line 1: should we name this assert_sample.yaml or something and have this common in all operators so that it can be referenced from install_yamls validation target?

cescgina: Done

tests/kuttl/common/assert_keystone_deployment.yaml
cescgina: Done. There is only one field that can't be obtained from the deployment, the `openshift.io/scc: anyuid` annotation that is only obtainable from the pod. I'm not sure how relevant that is for us, so I left a check for it, but can be removed if it's not that relevant.

tests/kuttl/common/assert_keystone_deployment.yaml
cescgina: Done. When checking the keystone service, I saw that there are actually 3: `keystone-admin`, `keystone-internal` and `keystone-public`, and the same is true for routes. I added checks for all three in services and routes. Also, I tried to add all the relevant fields, but I'm not 100% sure that I added all the needed ones, or whether all that I added are really relevant.

tests/kuttl/tests/keystone_scale/04-assert.yaml
```diff
@@ -0,0 +1,65 @@
...
+namespaced: True
+commands:
+  - script: |
+      set -x
+      available=$(oc get deployment -n openstack keystone -o yaml | grep "availableReplicas: 0")
+      if [ -z "$available"  ]; then
```
gibizer, line 61: as far as I understand this assert passes if `availableReplicas: 0` is not in the output. But this means it also passes if `availableReplicas: 1` is in the output. And what we want is that it passes if no `availableReplicas` is in the output. So you probably want to change `grep "availableReplicas: 0"` to `grep "availableReplicas"`.

//later

Or go with @stuggi's suggestion to use an error if there is any pods with `service: keystone ` label.

cescgina: You're right, thanks. I'm definitely checking the pods with an `errors`, it'll be cleaner

tests/kuttl/tests/keystone_scale/02-assert.yaml
cescgina: Done

tests/kuttl/tests/keystone_scale/03-assert.yaml
cescgina: Done

tests/kuttl/tests/keystone_scale/04-assert.yaml
cescgina: Moved the check to `04-errors.yaml` file

tests/kuttl/tests/keystone_scale/05-errors.yaml
cescgina: Done

tests/kuttl/tests/keystone_scale/05-errors.yaml
cescgina: Done

tests/kuttl/tests/keystone_scale/04-assert.yaml
cescgina: Moved the check to the `04-errors.yaml` file in the last commit

stuggi (approved):
thanks looks good to me to go to have this as a base to iterate on and use as an example for other service operators

PR #148, opened 2023-01-12
Remove admin token

kajinamit:
The admin token was deprecated a long ago when the new bootstrap process was added. The current implementation in keystone resources creation does not rely on it. We should avoid putting the unnecessary credential in config file.


stuggi (approved):
/lgtm

PR #151, opened 2023-01-18
Remove crc specific parameters from kuttl tests

cescgina:
Remove some fields from the asserts (host in routes and apiEndpoints in
CR spec) that had crc specific values since those make the tests fail if
run in a cluster other than crc. It should be easy to check with
a regex, but kuttl does not support it. If they're really important to
check, can introduce a TestAssert for them and check the value with some
script.

### DIFF
--- a/tests/kuttl/common/assert_sample_deployment.yaml
+++ b/tests/kuttl/common/assert_sample_deployment.yaml
@@ -235,3 +235,21 @@ spec:
   to:
     kind: Service
     name: keystone-public
+---
+# the actual addresses of the apiEndpoints are platform specific, so we can't rely on 
+# kuttl asserts to check them. This short script gathers the addresses and checks that
+# the three endpoints are defined and their addresses follow the default pattern
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: |
+      set -x
+      template='{{.status.apiEndpoint.admin}}{{":"}}{{.status.apiEndpoint.internal}}{{":"}}{{.status.apiEndpoint.public}}{{"\n"}}'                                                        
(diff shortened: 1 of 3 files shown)


stuggi (approved):
/lgtm

tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -43,10 +43,6 @@ spec:
       memory: 500Mi
   secret: osp-secret
 status:
-  apiEndpoint:
-    admin: http://keystone-admin-openstack.apps-crc.testing
-    internal: http://keystone-internal-openstack.apps-crc.testing
-    public: http://keystone-public-openstack.apps-crc.testing
```
gibizer, line 49: Unfortunately asserting that keystone has endpoints is part of the functionality we want to check here. So while I understand that the data here is deployment specific (ie. different in CRC than in SNO) simply not asserting such data feels wrong to me too. Unfortunately I cannot offer a solution here. Kuttl's assert possibilities seems limiting here.

stuggi: like run a command which uses a regexp?

cescgina: We could check with some script, it'll be less clean, but should be doable. Let me try and report back

gibizer: Yeah having a check that we have all the endpoints with some URL but ignoring the actual value of the URL would be better. I don't like having all these moving towards bash scripts as asserts but I don't see another option right now.

cescgina:
@stuggi @gibizer I've added a script to validate the endpoints urls. It assumes that the first part is always set to the same string (seems to be true for both CRC and SNO).

gibizer (approved):
Thanks

stuggi (approved):
/lgmt thanks!

PR #153, opened 2023-01-19
Add kuttl tests to check keystone config changes

cescgina:
Add kuttl test to check that changing the keystone service config
triggers a new pod with the new configuration. The tests check that
logging by default does not include debug information, then changes
the keystone config to set debug = true, and then checks that the logs
of the new pod now include debug information. As a last step, it checks
that the kesytone service is cleaned up.


tests/kuttl/common/scripts/check_keystone_pod_logs.sh
```diff
@@ -0,0 +1,20 @@
...
+pod=$(oc get pods -n openstack -l service=keystone -o name)
+# in the case the logs have DEBUG lines, keep only one to avoid cluttering the
+# test output
+debug=$(oc logs -n openstack "$pod" | grep "DEBUG" | head -n 1)
+
+if [ -n "$debug" ]; then
+    exit $found
+else
+    exit $not_found
+fi
```
abays, line 20: Perhaps naming this script more explicitly might be a good idea, i.e. `check_debug_in_keystone_pod_logs.sh`.  That way it indicates exactly what it checks the pod logs for.

abays (commented):
+1 from me, with one comment

tests/kuttl/common/scripts/check_keystone_pod_logs.sh
cescgina: Done, thanks.

tests/kuttl/common/errors_cleanup_keystone.yaml
```diff
@@ -0,0 +1,121 @@
...
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    admin: "true"
+    service: keystone
+  name: keystone-admin
+  namespace: openstack
+spec:
```
gibizer, line 41: Here and below: If we only want to assert that the Service is not exists then can we simply leave out the spec part? The metadata part should identify the service unambiguously.

tests/kuttl/tests/change_keystone_config/03-change_keystone_config.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      oc patch keystoneapi -n openstack keystone --type='json' -p='[{"op": "replace", "path": "/spec/defaultConfigOverwrite", "value":{"keystone.conf": "debug = true"}}]'
```
gibizer, line 5: I'm wondering if `spec.customServiceConfig` would be a better candidate for this test as the above command actually removes all the default configuration of the keystone service the operator creates and only leaves a keystone.conf with debug = True in it and that is not how we would like to enable debug logging in a real deployment.

gibizer (changes requested):

tests/kuttl/tests/change_keystone_config/02-assert.yaml
```diff
@@ -0,0 +1,6 @@
+# check that by default, debug is not set in keystone config
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: $KEYSTONE_KUTTL_DIR/../common/scripts/check_debug_in_keystone_pod_logs.sh --reverse
```
abays, line 6: Don't we deploy the `KeystoneAPI` sample with `debug = true` to start [1]?  Or are we doing some other manipulation to remove debug initially?

https://github.com/openstack-k8s-operators/keystone-operator/blob/d6671c8d0e22e4b6d4b4350223d54554a2980775/config/samples/keystone_v1beta1_keystoneapi.yaml#L10-L12

tests/kuttl/common/errors_cleanup_keystone.yaml
cescgina: Removed.

tests/kuttl/tests/change_keystone_config/03-change_keystone_config.yaml
cescgina: I tried to use `customServiceConfig` but I did not manage to get the configuration applied. Could you give me some indication on how to use it? Also, as @abays pointed out in another comment, the default value of  `customServiceConfig` already includes by default `debug = True`, but the logs do not show debug messages by default.

tests/kuttl/tests/change_keystone_config/02-assert.yaml
cescgina: Yes, and that is written to `/etc/keystone/keystone.conf.d/custom.conf`, but it seems that is not applied by default, since the logs do not show debug messages.

gibizer: I feel like you found some bugs :)

tests/kuttl/tests/change_keystone_config/03-change_keystone_config.yaml
gibizer: If the default is debug = True then either 
* a) you can deploy with a different yaml to have it not set by default
OR
* b) the test can be reversed. Check that there are DEBUG logs after deploy, set debug = False in customServiceConfig and check if the DEBUG logs does not appear any more.

tests/kuttl/tests/change_keystone_config/02-assert.yaml
abays: I think we're keeping everything in Jira, so let's add a bug there

tests/kuttl/tests/change_keystone_config/03-change_keystone_config.yaml
abays: I think the easiest thing is to do `b`.  

Currently in our testing, it does not look like the `spec.customServiceConfig` is working...but I think the problem might be on the OpenStack side as opposed to something we're doing wrong in the OCP operator.  I do see `/etc/keystone/keystone.conf.d/custom.conf` appearing in the Keystone pod with the expected `debug = true` setting under `[DEFAULT]`.  The Keystone service does not seem to be honoring it, however.

cescgina:
@gibizer @abays updated PR with the order of tests reversed. Debug is enabled by default, and the test checks that after disabling it using `customServiceConfig` it's not present in the logs. It needs https://github.com/openstack-k8s-operators/keystone-operator/pull/201 to be merged to work

cescgina:
@gibizer @abays After merging #201 and #204 tests are passing and this PR is ready again for review

abays (approved):
/lgtm

PR #161, opened 2023-01-27
Assign the service role to service users

kajinamit:
... in addition to the admin role. The service role is required for the following features.

- Some components such as nova uses service user token to interact with a different component with avoiding token expiration. Currently this feature can be used by any roles but in the future this would be allowed for only users with the service role.

- Keystone allows only system readers, service users or token subject to verify a token when SRBAC is enforced in Keystone.


kajinamit:
See
 https://review.opendev.org/c/openstack/tripleo-heat-templates/+/703712
and
 https://review.opendev.org/c/openstack/tripleo-heat-templates/+/819250
for reference. In recent version of TripleO we assign the service role to all service users.

stuggi (approved):
/lgtm

PR #162, opened 2023-01-27
Add webhook scaffold

rabi:
Add validating webhook and changes to selector labels to not conflict with other installed operators.


api/bases/keystone.openstack.org_keystoneapis.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.9.2
```
abays, line 6: It looks like your environment is using an older Golang/Operator-SDK than what we've recently upgraded to

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: quay.io/rabi/keystone-operator
```
abays, line 15: This should be `quay.io/openstack-k8s-operators/keystone-operator`, or at least shouldn't be your own repo for now

main.go
```diff
@@ -124,6 +124,10 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "KeystoneEndpoint")
 		os.Exit(1)
 	}
+	if err = (&keystonev1.KeystoneAPI{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "KeystoneAPI")
+		os.Exit(1)
+	}
```
abays, line 130: I think we should also add the self-signed cert `make` target like you did for the similar Glance webhook PR

api/bases/keystone.openstack.org_keystoneapis.yaml
rabi: Looks like it uses old controller-gen in ./bin unless you clean it  and overwrites things. I'll see if I can fix it in another PR.

main.go
rabi: Yeah, I've not updated it waiting for the glance one to merge.

api/bases/keystone.openstack.org_keystoneapis.yaml
rabi: https://github.com/openstack-k8s-operators/keystone-operator/pull/179

main.go
rabi: Done!

abays:
@rabi Could we rebase this?  I'd like to merge it and then I will do a follow-up PR to massage the default images pattern into Keystone (which uses the webhooks).

rabi:
> @rabi Could we rebase this? I'd like to merge it and then I will do a follow-up PR to massage the default images pattern into Keystone (which uses the webhooks).

@abays Looks like your patch has already been merged. I assume we don't need this one.

PR #163, opened 2023-01-30
Adapt assert check to use with sh and not bash

cescgina:
When using TestAsserts in kuttl tests, the scripts are passed to 'sh -c', which  may or may not be a bash shell. To avoid failures in cases where  the shell is actually not bash, this change removes any bash-specific features from the kuttl tests.

### DIFF
--- a/tests/kuttl/common/assert_sample_deployment.yaml
+++ b/tests/kuttl/common/assert_sample_deployment.yaml
@@ -244,11 +244,11 @@ kind: TestAssert
 namespaced: true
 commands:
   - script: |
-      set -x
       template='{{.status.apiEndpoint.admin}}{{":"}}{{.status.apiEndpoint.internal}}{{":"}}{{.status.apiEndpoint.public}}{{"\n"}}'
       regex="http:\/\/keystone-admin-openstack\.apps.*:http:\/\/keystone-internal-openstack\.apps.*:http:\/\/keystone-public-openstack\.apps.*"
-      apiEndpoints=$(oc get -n openstack KeystoneAPI  keystone -o go-template=$template)
-      if [[ "$apiEndpoints" =~ $regex ]]; then
+      apiEndpoints=$(oc get -n openstack KeystoneAPI  keystone -o go-template="$template")
+      matches=$(echo "$apiEndpoints" | sed -e "s?$regex??")
+      if [ -z "$matches" ]; then
         exit 0
       else
         exit 1


gibizer (changes requested):
Isn't `set -x` bash specific as well? 

I suggest to document this rule (no bash specific in script part of kuttl) in https://github.com/openstack-k8s-operators/docs 

Also it would be nice to run https://github.com/koalaman/shellcheck on the script parts of the kuttl test code. I think that can also catch bashism in this scripts so we can avoid merging them.

cescgina:
> Isn't `set -x` bash specific as well?
That's true, removed.
> I suggest to document this rule (no bash specific in script part of kuttl) in https://github.com/openstack-k8s-operators/docs
Didn't know about that repo, I'll add it there.
> Also it would be nice to run https://github.com/koalaman/shellcheck on the script parts of the kuttl test code. I think that can also catch bashism in this scripts so we can avoid merging them.
I'm looking into whether shellcheck can check only part of a file, but I run it manually and it caught a couple of possible problems that I fixed. Thanks for the suggestion.

gibizer:
> > Isn't `set -x` bash specific as well?
> 
> That's true, removed.
> 
> > I suggest to document this rule (no bash specific in script part of kuttl) in https://github.com/openstack-k8s-operators/docs
> 
> Didn't know about that repo, I'll add it there.

Thanks
> 
> > Also it would be nice to run https://github.com/koalaman/shellcheck on the script parts of the kuttl test code. I think that can also catch bashism in this scripts so we can avoid merging them.
> 
> I'm looking into whether shellcheck can check only part of a file, but I run it manually and it caught a couple of possible problems that I fixed. Thanks for the suggestion.

You can grab the script part of the yaml file and pass it to shellcheck:
```
yq -r  '.commands[]?.script' tests/kuttl/common/assert_sample_deployment.yaml | shellcheck -
```
we can even include some make target / CI job to lint all the kuttl file with
```
find ./tests/kuttl/  -name '*.yaml' | xargs yq -r  '.commands[]?.script' | shellcheck -eSC2317 -
```
I ignored SC2317 as it seems to be a false positive in our case where the scripts are called indirectly via kuttl.

gibizer (approved):

stuggi (approved):
/lgtm

cescgina:
@stuggi sorry, I didn't realize that the PR needed a rebase, could you please re-apply the lgtm label?

PR #168, opened 2023-01-31
Support adopting KeystoneService from existing DB

gibizer:
During DB adoption we are copying over the keystone DB schema from an existing DB server. If that schema already contains services then the keystone_service controller needs to make sure that the ServiceID of a given service is propagated from the DB to the
KeystoneService.Status.ServiceID field when the KeystoneService CR is created. Without that the KeystoneService won't get into Ready state.

See the discussion about the adoption issue in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/3

### DIFF
--- a/controllers/keystoneservice_controller.go
+++ b/controllers/keystoneservice_controller.go
@@ -370,20 +370,27 @@ func (r *KeystoneServiceReconciler) reconcileService(
 		if err != nil {
 			return err
 		}
-	} else if service.Enabled != instance.Spec.Enabled ||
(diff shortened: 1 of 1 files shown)


gibizer:
I keep this as a draft while I test this to see if it full resolves the adoption issue

gibizer:
I tested this successfully. So with this patch the placement KeystoneEndpoint is created successfully during adoption as the KeystoneService now has ServiceID imported from the adopted DB.

stuggi (approved):
/lgtm

PR #174, opened 2023-02-01
Improve Service/Endpoint deletion logic

abays:
An approach to preventing finalizer lock-up for the Keystone operator in certain situations where you try to delete a `KeystoneService` or `KeystoneEndpoint` before the `KeystoneAPI` has come up completely

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -611,10 +611,8 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
 	return ctrl.Result{}, nil
 }
 
-//
 // generateServiceConfigMaps - create create configmaps which hold scripts and service configuration
 // TODO add DefaultConfigOverwrite
-//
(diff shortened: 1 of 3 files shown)


gibizer (commented):
This is getting fairly complicated now that we have 3 different ways to end up in `reconcileDelete`. I played with couple of ideas locally to move the are-we-being-deleted check around to see if it helps unifying cases but I failed finding a simpler solution.

abays:
> This is getting fairly complicated now that we have 3 different ways to end up in `reconcileDelete`. I played with couple of ideas locally to move the are-we-being-deleted check around to see if it helps unifying cases but I failed finding a simpler solution.

I agree with you -- this is clunky and I don't necessarily like it.  We have to use finalizers if we want to protect resources from improper deletion, but that unfortunately walks us into situations like this where the operator pattern we've applied begins to become a bit convoluted.

gibizer (approved):
The scenario is real as KeystoneEndpoint and KeystoneService CR can be created before KeystoneAPI is Ready so delete should also work in that case. I have failed to come up with a simpler solution. So lets merge this.

PR #175, opened 2023-02-01
Support exposing service via metallb

stuggi:
* renames NetworkAttachmentDefinitions to NetworkAttachments
* introduce externalEndpoints Parameters to configure metallb service using the pre-created metallb IPAddressPool and L2Advertisement . 

E.g. to expose the api on an internalapi network which
* should be registered as the internal endpoint
* its metallb ipaddresspool is configured in an osp-internalapi object
*  and request a specific loadbalancerIP which is shared with other services (list of IPs for multi stack ipv4/ipv6)

```
...
spec:
...
      externalEndpoints:
      - endpoint: internal
        loadbalancerIPs:
        - 172.17.0.202
        ipAddressPool: osp-internalapi
        sharedIP: true
```

* removes registration of admin endpoint

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/160


gibizer:
The code looks good to me but I have limited knowledge around metallb.

controllers/keystoneapi_controller.go
```diff
@@ -537,8 +547,37 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
...
+				err.Error()))
+			return ctrl.Result{}, err
+		}
+	}
+
+	serviceAnnotations, err := nad.CreateNetworksAnnotation(instance.Namespace, instance.Spec.NetworkAttachments)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed create network annotation from %s: %w",
+			instance.Spec.NetworkAttachments, err)
+	}
```
abays, line 577: I wonder if eventually we might want to surface this in the conditions, but I suppose it will be rare for that `nad.CreateNetworksAnnotation(...)` func to ever return an error.

gibizer (commented):
looks good to me

olliewalsh (approved):
/lgtm

PR #176, opened 2023-02-02
Fix pattern to exclude bin

kajinamit:
Current pattern can match subdirectories.

Also, add the testbin directory to make the file consistent with the other operators.


abays (approved):
/lgtm

PR #177, opened 2023-02-02
Fix wrong secret name in error log

kajinamit:
We should put the secret name instead of the global secret name.


abays (approved):
/lgtm

PR #178, opened 2023-02-02
Generate keys for credentials

kajinamit:
Keystone uses fernet mechanism for not only tokens but also credentials and requires key files to use these features.

This introduces an additional secret file to store credential keys.


kajinamit:
Tested in my local env. Note that this requires redeploying the `keystoneapis` instance or recreating the `keystone` secret.

kajinamit:
We might want to revisit the previous version which generates secrets for token and credential separately. Current version does not allow us to add additional key without modifying the code.

stuggi (approved):
/lgtm

PR #179, opened 2023-02-03
Cleanup old binaries

rabi:
There could be older version of the binaries that would mess up things.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,17 +187,19 @@ KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/k
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
 	rm -f $(LOCALBIN)/kustomize || true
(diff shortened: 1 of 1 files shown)


dprince:
would this cause the binaries to be downloaded each time though?

rabi:
> would this cause the binaries to be downloaded each time though?

Yes. We can check for the existing version before downloading though. However, we already clean/download for other binaries[1] and probably is not a big overhead.


[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/master/Makefile#L189

dprince:
> > would this cause the binaries to be downloaded each time though?
> 
> Yes. We can check for the existing version before downloading though. However, we already clean/download for other binaries[1] and probably is not a big overhead.
> 
> [1] https://github.com/openstack-k8s-operators/keystone-operator/blob/master/Makefile#L189

I'm not sure I agree we should be doing that. (we should remove the cleaning of kustomize each time!)

Rather than cleaning these each time, and forcing users to download could we could add a cleanup Makefile target and let folks call that if they wish.

rabi:
> Rather than cleaning these each time, and forcing users to download could we could add a cleanup Makefile target and let folks call that if they wish.

Well, as I mentioned earlier user won't even realize that they have an old binary and would 'whish' to download. PRs would be changed to use old one[1]. But I can't argue enough about this inconvenience.

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/162#discussion_r1094328439

PR #185, opened 2023-02-06
Fix wrong error handling of EnsureSecrets

kajinamit:
The EnsureSecrets function returns nil if the secret exists or is created and returns error if something goes wrong.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -794,11 +794,10 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
 			},
 		}
 		err := oko_secret.EnsureSecrets(ctx, helper, instance, tmpl, envVars)
-		if err != nil {
-			return nil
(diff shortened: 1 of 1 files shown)


controllers/keystoneapi_controller.go
```diff
@@ -794,11 +794,10 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
 			},
 		}
 		err := oko_secret.EnsureSecrets(ctx, helper, instance, tmpl, envVars)
+
 		if err != nil {
-			return nil
+			return fmt.Errorf("OpenStack secret %s not found", secretName)
```
rabi, line 799: Why override the error coming from lib-common? just 'return err' would be better I think.

kajinamit: totally agree. Updated the change accordingly.

abays (approved):
/lgtm

PR #187, opened 2023-02-10
Return error in ConfigMap creation

kajinamit:
The current logic ignores the error but it should be returned so that the error can be detected in the reconciler.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -661,12 +661,7 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
 			Labels:        cmLabels,
 		},
 	}
-	err := configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
-	if err != nil {
-		return nil
-	}
-
-	return nil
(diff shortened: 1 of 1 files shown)


controllers/keystoneapi_controller.go
```diff
@@ -661,12 +661,7 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
 			Labels:        cmLabels,
 		},
 	}
-	err := configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
-	if err != nil {
-		return nil
-	}
-
-	return nil
+	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
```
marios, line 664: adding comment in case useful for anyone else

i was just trying to understand and it took a few minutes to find that EnsureConfigMaps - defined there https://github.com/openstack-k8s-operators/lib-common/blob/f7b52e2a2ccbffb44912163f9d072eda699e781b/modules/common/configmap/configmap.go#L159

abays (approved):
/lgtm

Issue #188, opened 2023-02-10
Redeploying keystone terminates controller pod with nil pointer dereference


rabi:
$ make keystone_deploy
$ oc delete keystoneapi/keystone
$ make keystone_deploy

controller taceback:

r": "keystoneapi", "controllerGroup": "keystone.openstack.org", "controllerKind": "KeystoneAPI", "KeystoneAPI": {"name":"keystone","namespace":"openstack"}, "namespace": "openstack", "name": "keystone", "reconcileID": "0e0885c2-9990-4f85-bbca-4643cd3abb3c"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x16184a8]

goroutine 660 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/home/stack/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:119 +0x1fa
panic({0x17e3800, 0x28e9870})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/openstack-k8s-operators/keystone-operator/controllers.(*KeystoneAPIReconciler).ensureFernetKeys(0xc000a99500, {0x1c8d850, 0xc0025b5e60}, 0xc0025c2280, 0x5f?, 0xc00035e670)
	/home/stack/work/keystone-operator/controllers/keystoneapi_controller.go:811 +0x448
github.com/openstack-k8s-operators/keystone-operator/controllers.(*KeystoneAPIReconciler).reconcileNormal(0xc000a99500, {0x1c8d850, 0xc0025b5e60}, 0xc0025c2280, 0xc0025c2280?)
	/home/stack/work/keystone-operator/controllers/keystoneapi_controller.go:507 +0x625
github.com/openstack-k8s-operators/keystone-operator/controllers.(*KeystoneAPIReconciler).Reconcile(0xc000a99500, {0x1c8d850, 0xc0025b5e60}, {{{0xc0026b1490, 0x9}, {0xc0026b1480, 0x8}}})
	/home/stack/work/keystone-operator/controllers/keystoneapi_controller.go:175 +0xafa
[...]

PR #189, opened 2023-02-10
Fix esnureFernetKeys()

rabi:
secret is returned as nil from GetSecret() if the secret is not found. However, we try to use secret.Name before returning. This results in nil pointer derefence error and kills the controller.

Closes: #188

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -799,12 +799,12 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
 		if err != nil {
 			return err
 		}
+	} else {
+		// add hash to envVars
+		(*envVars)[secret.Name] = env.SetValue(hash)
 	}
 
 	// TODO: fernet key rotation
-
-	// add hash to envVars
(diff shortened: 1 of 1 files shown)


kajinamit (approved):

kajinamit:
Or alternatively should we get the secret hash after it is created ?

rabi:
> Or alternatively should we get the secret hash after it is created ?

lib-common could be refactored to make the api better, but at present we populate the hash already in EnsureSecrets()[1]  after creating it.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go#L287

kajinamit:
> > Or alternatively should we get the secret hash after it is created ?
> 
> lib-common could be refactored to make the api better, but at present we populate the hash already in EnsureSecrets()[1] after creating it.
> 
> [1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go#L287

Ah, ok. I missed that. totally makes sense.

stuggi (approved):
/lgtm

PR #200, opened 2023-02-21
Add omitempty tag to nillable fields

rabi:
Otherwise marshalling/unmarshalling CRDs with webhooks results in error
as it serializes these struct fields as "null".

`The KeystoneAPI "keystone" is invalid: spec.networkAttachmentDefinitions:
Invalid value: "null": spec.networkAttachmentDefinitions in body must be of
type array: "null"`

### DIFF
--- a/api/v1beta1/keystoneapi_types.go
+++ b/api/v1beta1/keystoneapi_types.go
@@ -128,16 +128,16 @@ type KeystoneAPISpec struct {
 
 	// +kubebuilder:validation:Optional
 	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
-	NetworkAttachments []string `json:"networkAttachments"`
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
 
 	// +kubebuilder:validation:Optional
 	// ExternalEndpoints, expose a VIP using a pre-created IPAddressPool
-	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints"`
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
 }
 
 // MetalLBConfig to configure the MetalLB loadbalancer service
 type MetalLBConfig struct {
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=internal;public
 	// Endpoint, OpenStack endpoint this service maps to
 	Endpoint endpoint.Endpoint `json:"endpoint"`
@@ -161,7 +161,7 @@ type MetalLBConfig struct {
 
 	// +kubebuilder:validation:Optional
 	// LoadBalancerIPs, request given IPs from the pool if available. Using a list to allow dual stack (IPv4/IPv6) support
-	LoadBalancerIPs []string `json:"loadBalancerIPs"`
+	LoadBalancerIPs []string `json:"loadBalancerIPs,omitempty"`
 }
 
 // PasswordSelector to identify the DB and AdminUser password from the Secret
--- a/api/bases/keystone.openstack.org_keystoneapis.yaml
+++ b/api/bases/keystone.openstack.org_keystoneapis.yaml
@@ -147,6 +147,7 @@ spec:
                         specified.
                       type: string
                   required:
+                  - endpoint
                   - ipAddressPool
                   type: object
                 type: array
--- a/config/crd/bases/keystone.openstack.org_keystoneapis.yaml
+++ b/config/crd/bases/keystone.openstack.org_keystoneapis.yaml
@@ -147,6 +147,7 @@ spec:
                         specified.
                       type: string
                   required:
+                  - endpoint
                   - ipAddressPool
                   type: object
                 type: array


api/v1beta1/keystoneapi_types.go
```diff
@@ -128,19 +128,19 @@ type KeystoneAPISpec struct {
 
 	// +kubebuilder:validation:Optional
 	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
-	NetworkAttachments []string `json:"networkAttachments"`
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
```
gibizer, line 131: I'm wondering, would a `// +kubebuilder:default=[]` solve this too? If so then I would go that way as that would actually document to the API user what will be the value of the field if it is not provided.

api/v1beta1/keystoneapi_types.go
```diff
@@ -128,19 +128,19 @@ type KeystoneAPISpec struct {
...
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
 }
 
 // MetalLBConfig to configure the MetalLB loadbalancer service
 type MetalLBConfig struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Enum=internal;public
 	// Endpoint, OpenStack endpoint this service maps to
-	Endpoint endpoint.Endpoint `json:"endpoint"`
+	Endpoint endpoint.Endpoint `json:"endpoint,omitempty"`
```
gibizer, line 143: This is a strange one.  Optional would allow empty value which means "" for the golang string behind Endpoint, but the Enum validation rejects that value. So this is not really an optional field. I suggest to make this Required.

gibizer (dismissed):

stuggi (approved):
/lgtm

api/v1beta1/keystoneapi_types.go
stuggi: for keystone itself yes, but when we move this struct to the infra-operator later, its optional with the intention that you could expose a service via a vip which is not used/registered as an endpoint

stuggi:
> /lgtm

removed it as it clashed with Gibi's comments, questions

api/v1beta1/keystoneapi_types.go
gibizer: What will be the value of the Endpoint golang struct field in that case? The current definition only allow two possible values `internal` and `public`.

stuggi: e.g. if the DB is ever needed to be exposed

stuggi: > What will be the value of the Endpoint golang struct field in that case? The current definition only allow two possible values `internal` and `public`.

you don't set an endpoint in that case where you don't register the vip as an endpoint. the alternative is we introduce a `none` value.

gibizer: Changing the validation to `// +kubebuilder:validation:Enum=internal;public;none`  would allow expressing if no Endpoint is needed. Still in that case I would make the defintion `// +kubebuilder:validation:Required` to avoid confusion as it always need one of the 3 possible values.

stuggi: right, if we introduce `none`, it can be required

stuggi: I'd suggest that we make this required and keep the current two values. Right now there is no need for the use case I described. We'll introduce that as a follow up as it also requires some work in lib-common.

api/v1beta1/keystoneapi_types.go
stuggi: works for me if that is ok for the work @rabi is doing right now.

api/v1beta1/keystoneapi_types.go
gibizer: The whole optional, non defaulted CRD field + golang struct unmarshalling is problematic. Golang defines its own set of default values if the field is uninitialized. In this case the string endpoint field will be initialized to the "" string if the json request does not contain the field. But the marshaling of the golang struct back to json will pick up the """ string as a value for the json representation instead of skipping the field in the json output. This is where omitempty cames to the picture instructing the json emitter to skip the whole field if the value of the field happen to be the empty golang default value.

We saw in the past that using omitempty in a Spec field is problematic as it prevents forcing a the field value to the golang default if there is another value set before. See https://github.com/openstack-k8s-operators/nova-operator/pull/177

rabi: Using defaults would probably work from serialization/deserialization pov for arrays, maps and struct and even fields, but, I think this has to change across all operator CRDs. We currently use omitempty everywhere including string fields[1] that are optional.

I think the current assumption is if a field is optional it can be omitted. IMO, we should not change that for a few newly added fields only.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/master/api/v1beta1/keystoneapi_types.go#L173
https://github.com/openstack-k8s-operators/keystone-operator/blob/master/api/v1beta1/keystoneapi_types.go#L122

rabi: I also feel that making it optional and providing a empty default is kind of counter intuitive.

rabi: However, I'll make the change suggested to get me unblocked atm. But we probably need to be consistent across all CRDs and have some of these assumptions documented.

api/v1beta1/keystoneapi_types.go
```diff
@@ -127,17 +127,18 @@ type KeystoneAPISpec struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=[]
 	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
 	NetworkAttachments []string `json:"networkAttachments"`
 
 	// +kubebuilder:validation:Optional
```
stuggi, line 134: lets default this as well `// +kubebuilder:default=[]`

rabi: Looks like the default does not work as it's treated as string?

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_placement-operator/133/pull-ci-openstack-k8s-operators-placement-operator-master-functional/1627975037304705024

abays:
> Otherwise marshalling/unmarshalling CRDs with webhooks results in error as it serializes these struct fields as "null".
> 
> `The KeystoneAPI "keystone" is invalid: spec.networkAttachmentDefinitions: Invalid value: "null": spec.networkAttachmentDefinitions in body must be of type array: "null"`

I am hitting the same problem with some Glance webhook stuff I am doing:

```
The Glance "glance" is invalid: 
* spec.extraMounts: Invalid value: "null": spec.extraMounts in body must be of type array: "null"
* spec.glanceAPIExternal.externalEndpoints: Invalid value: "null": spec.glanceAPIExternal.externalEndpoints in body must be of type array: "null"
...
etc
```

api/v1beta1/keystoneapi_types.go
gibizer: OK I think this was my mistake the proper default for empty slice is `{}`, so ` // +kubebuilder:default={}`

See: https://book.kubebuilder.io/reference/markers.html

abays (approved):
I guess since `omitempty` seems to be what is required for proper marshalling/unmarshalling in mutating webhooks, this is good for now

/lgtm

PR #201, opened 2023-02-21
Remove log_config_append from keystone config

cescgina:
The log_config_append line in the template keystone.conf seems to
interfere with normal config handling and prevents debug settings
introduced from keystone.conf.d/custom.conf which are sent through the
customServiceConfig field in KeystoneAPI. If the log_config_append
option is set in the keystone.conf template is set and debug = true is
set in custom.conf, then the custom.conf config is ignored and no debug
logs are shown.


abays:
Could this cause any side effects we are not anticipating?

cescgina:
> Could this cause any side effects we are not anticipating?

Logs look normal, what we had before + DEBUG logs, but I don't know enough about keystone config to know for sure.

dprince:
I think we could probably remove this from the config by we probably want to replace it with something that logs to stdout right? It looks like including /etc/keystone/logging.conf setups default *file* logging.

rabi (changes requested):
Yeah, we don't need that as default (which is to redirect to a separate logging configuration file) . But there are more changes required. Please check https://github.com/openstack-k8s-operators/ironic-operator/commit/d9402f4dd3a4343a1fb32200d05153012788bceb

cescgina:
> Yeah, we don't need that as default (which is to redirect to a separate logging configuration file) . But there are more changes required. Please check [openstack-k8s-operators/ironic-operator@d9402f4](https://github.com/openstack-k8s-operators/ironic-operator/commit/d9402f4dd3a4343a1fb32200d05153012788bceb)

Thanks @rabi, I've updated the PR imitating the change you linked. Let me know if I missed anything.

abays (approved):
Looks like @rabi 's requested changes have been made

/lgtm

rabi (approved):

PR #203, opened 2023-02-23
Set owner references for configmaps

rabi:
This ensures that they're garbage collected.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -660,7 +660,7 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
 	//
 	// create OpenStackClient config
 	//
-	err = r.reconcileConfigMap(ctx, instance)
+	err = r.reconcileCloudConfig(ctx, helper, instance)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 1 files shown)


controllers/keystoneapi_controller.go
```diff
@@ -723,7 +723,10 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
 
 // reconcileConfigMap -  creates clouds.yaml
 // TODO: most likely should be part of the higher openstack operator
-func (r *KeystoneAPIReconciler) reconcileConfigMap(ctx context.Context, instance *keystonev1.KeystoneAPI) error {
+func (r *KeystoneAPIReconciler) reconcileConfigMap(
```
abays, line 726: Should this function instead use `lib-common`'s config map functions like the `KeystoneAPI` controller already is here: https://github.com/openstack-k8s-operators/keystone-operator/blob/d3b76057f74302810cf3e299421f9785a3ac3491/controllers/keystoneapi_controller.go#L674-L722?

rabi: Probably we should. maybe in a follow-up(?). I keep bug fixes and refactoring separate so as not to introduce more bugs.

rabi: I've updated it to use lib-common instead. The CI failure does not seem related.

rabi: CI issue resolved.

abays (approved):
/lgtm

PR #204, opened 2023-02-28
Adjust kuttl test for changes in endpoints

cescgina:
Recently the admin endpoint was removed, along with the service and
route, this changes removes them from the kuttl assert.


stuggi (approved):
/lgtm

PR #205, opened 2023-03-01
Sync CRD manifests

gibizer:
It seems the stored CRD manifests are not in sync with the golang type definition and it makes the pre-commit job fail. This commit sync the manifests by running `make manifests` and stores the result in the repo.


dprince:
@gibizer pre-commit seems to run fine for me locally though? How can I reproduce what you are fixing here?

PR #207, opened 2023-03-07
Add cronjob to purge expired trusts

kajinamit:
This change introduces the CronJob resource to periodically purge expired trusts from database.

### DIFF
--- a/api/v1beta1/keystoneapi_types.go
+++ b/api/v1beta1/keystoneapi_types.go
@@ -89,6 +89,21 @@ type KeystoneAPISpec struct {
 	Secret string `json:"secret"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	// TrustFlushArgs - Arguments added to keystone-manage trust_flush command
+	TrustFlushArgs string `json:"trustFlushArgs"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="1 * * * *"
+	// TrustFlushSchedule - Schedule to purge expired or soft-deleted trusts from database
+	TrustFlushSchedule string `json:"trustFlushSchedule"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// TrustFlushSuspend - Suspend the cron job to purge trusts
+	TrustFlushSuspend bool `json:"trustFlushSuspend"`
+
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:default={database: KeystoneDatabasePassword, admin: AdminPassword}
 	// PasswordSelectors - Selectors to identify the DB and AdminUser password from the Secret
 	PasswordSelectors PasswordSelector `json:"passwordSelectors"`
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	configmap "github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
+	cronjob "github.com/openstack-k8s-operators/lib-common/modules/common/cronjob"
 	deployment "github.com/openstack-k8s-operators/lib-common/modules/common/deployment"
 	endpoint "github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
 	env "github.com/openstack-k8s-operators/lib-common/modules/common/env"
@@ -90,6 +91,7 @@ type KeystoneAPIReconciler struct {
 // +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete;
+// +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=route.openshift.io,resources=routes,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbdatabases,verbs=get;list;watch;create;update;patch;delete;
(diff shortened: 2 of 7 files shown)


pkg/keystone/cronjob.go
```diff
@@ -0,0 +1,125 @@
...
+	parallelism := int32(1)
+	completions := int32(1)
+
+	cronjob := &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ServiceName + "-cron",
+			Namespace: instance.Namespace,
+		},
+		Spec: batchv1.CronJobSpec{
+			Schedule:          "1 * * * *",
```
stuggi, line 62: I think that's a good value, just wondering if we ever need to expose the `schedule` for cronjobs? maybe not for this one, just a general thought.

pkg/keystone/cronjob.go
```diff
@@ -0,0 +1,125 @@
...
+	// If possible two pods of the same service should not
+	// run on the same worker node. If this is not possible
+	// the get still created on the same worker node.
+	cronjob.Spec.JobTemplate.Spec.Template.Spec.Affinity = affinity.DistributePods(
+		common.AppSelector,
+		[]string{
+			ServiceName + "-cron",
+		},
+		corev1.LabelHostname,
+	)
```
stuggi, line 107: is affinity required if the Parallelism is 1?

kajinamit: No. I'll remove this.

Just for record.
We generally do not require multiple pods to run cron jobs. The only exception would be one used by glance which should be run at the same node(with /var/lib/glance) shared with glance-api but that don't require affinity either.

pkg/keystone/cronjob.go
kajinamit: Yeah it makes sense to expose the schedule. Probably I can expose this in keystone-operator as well, so that users can tune frequency of the purge job.

In the past we provided an option to disable the cron job but I'm not sure that is required. We probably can omit it in operators until we hear real requirements.

Finally, I noticed I have to provide the interface to customize arguments of that command (eg ``--date``). I'll add that.

kajinamit: I've added the parameter to expose the schedule and additional command arguments.

controllers/keystoneapi_controller.go
```diff
@@ -654,9 +657,24 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
...
+	// create CronJob
+	cronjobDef := keystone.CronJob(instance, serviceLabels, serviceAnnotations)
+	cronjob := cronjob.NewCronJob(
+		cronjobDef,
+		time.Duration(5)*time.Second,
+	)
+
+	ctrlResult, err = cronjob.CreateOrPatch(ctx, helper)
+	if err != nil {
+		// TODO: Do we need a different status ?
```
bshephar, line 671: I guess the question is whether we log this action anywhere else? If we run it from within a job pod, it probably won't be writing to the normal service log files write? So if a user (or someone supporting the environment) needed to check the last time `trust_flush` was executed, if they can't get that information from a log file. Then it would probably be useful to report something in the status field when the job succeeds or fails?

pkg/keystone/cronjob.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	envVars := map[string]env.Setter{}
+	envVars["KOLLA_CONFIG_FILE"] = env.SetValue(KollaConfig)
+	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
+
+	parallelism := int32(1)
+	completions := int32(1)
+
+	cronjob := &batchv1.CronJob{
```
bshephar, line 55: If we set `spec.successfulJobsHistoryLimit` here, we might not need to worry about adding a status condition for it? We could keep say the last n jobs?

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#jobs-history-limits

Then a user just needs to check `oc get jobs` and the execution to see if it worked or not?

kajinamit: In short, no. The concern left in keystoneapi_controller.got is about the way how we handle failure while creating/patching a CronJob resource and is not related to how actual jobs are executed. In case the operatro failed to create a cron job then there are no cronjobs and jobs created.

controllers/keystoneapi_controller.go
kajinamit: There are two levels of failures we have to consider
- Failure while creating a CronJob resource
- Failure in actual jobs triggered by CronJob
and operator would be responsible with only one and checking actual jobs would be out of its scope.

I have to double-check this but in my understanding the error is recorded in operator logs if the operator fails to create a cron job. The question is more about whether we should expose the failure as part of CR status.

bshephar: I think it's probably worthwhile setting a status.condition for it. Just for convenience sake, if it fails to create the cronjob it could go unnoticed by users if it's just hidden in the logs of the operator. imo, it would be worth adding a status.condition for it. But happy to wait for opinions from others as well.

stuggi: yes, we should have a  condition for the creating the CronJob resource it's part of setting up the service with its dependencies. 

If we can get the status, maybe just run time and successful/failed of the last run, also in a separate condition, this would make the condition the first spot to start with debugging. But then we should exclude this last run condition for calculate the overall Ready state of the service.

stuggi: re the overall Ready state, at least until it was running for the first time and as not an unknown state.

kajinamit: I'd be careful with adding logic to check succesfull run because it might take long until the job is triggered for the first time(eg. if the job is executed once a day then it might take 24 hours in the worst case until the job is run.)

I think we can start with operators taking care of resource creation and let users to use cronjob/job interfaces directly to check how jobs are executed.

stuggi: makes sense

api/v1beta1/keystoneapi_types.go
```diff
@@ -89,6 +89,16 @@ type KeystoneAPISpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	// TrustFlushArgs - Arguments added to keystone-manage trust_flush command
+	TrustFlushArgs string `json:"trustFlushArgs"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="1 * * * *"
+	// TrustFlushSchedule - Schedule to purge expired or soft-deleted trusts from database
+	TrustFlushSchedule string `json:"trustFlushSchedule"`
+
```
stuggi, line 101: do we need to be able to control suspend of the job, or is setting a really long schedule the alternative to not run it for the time being?

stuggi: https://kubernetes.io/blog/2021/04/12/introducing-suspended-jobs/

controllers/keystoneapi_controller.go
kajinamit: Now the latest version should update the status according to result of cronjob resource creation.

api/v1beta1/keystoneapi_types.go
kajinamit: I think we can't use that suspend interface because jobs are created by cronjob controller. I checked the current interface of CronJob resource but could not find the way to disable jobs, We probably should remove the CronJob resource in case we want to disable job executuins.

kajinamit:
/retest

I've included change of default images unexpectedly and removed it. seems prow jobs still pulled the previous version.

kajinamit:
Hmm I'm not too sure what is causing the failure in kuttl test but I've deployed this code in my local env and confirmed the resources are created and all status conditions are set to True.

kajinamit:
Thanks @abays ! It seems the job is now passing now.

api/v1beta1/keystoneapi_types.go
kajinamit: I noticed I was looking at something wrong aid CronJob does support suspend. I've updated the change to support it.

stuggi (approved):
/lgtm

PR #213, opened 2023-03-08
Do not run redundant github actions running via prow

stuggi:
The github actions govet, lint, gotest, golangci ... run now on PRs via prow, which is redundant. Lets only run them on push, as its still useful to run them when pushing to the fork.


bshephar (approved):
We should probably apply this globally. We have a few redundancies in our Github / Prow jobs  across the board I believe, even beyond this change I mean. One example is that we often run envtest with GitHub actions + again in Prow with functional tests.

PR #216, opened 2023-03-13
Fix omitempty + default value

gibizer:
Having a CRD field with omitempty tag with a kubebuilder default value defined can cause unexpected behavior. See
https://github.com/gibizer/operator-lint/blob/main/linters/crd/C003 for details.

So this patch remove the problematic definitions and bumps the operator-lint version to 0.3.0 that has a check to cover this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,5 +316,5 @@ gowork: ## Generate go.work file to support our multi module repository
 
 .PHONY: operator-lint
 operator-lint: gowork ## Runs operator-lint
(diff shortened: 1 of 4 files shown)


gibizer:
/test precommit-check

golangci-lint timed out, we probably needs to bump the timeout value eventually

gibizer:
We need https://github.com/openstack-k8s-operators/keystone-operator/pull/217 to increase the timeout of golangci-lint run from 1 min to 5 min.

stuggi (approved):
/lgtm

PR #217, opened 2023-03-13
Increase the timeout of golangci-lint

gibizer:
We see pre-commit job timeouts in CI due to golangci-lint running longer than the default 1 minute. So this PR bumping the timeout to 5 mins and enables verbose logging to see what take much time there.


stuggi (approved):
/lgtm

PR #224, opened 2023-03-29
Remove install_yamls dependency for kuttl tests

cescgina:
Instead of using install_yamls targets for deploying and cleaning up
keystone, use the CR sample from config/samples. This removes a possible
circular dependency issue when using install_yamls to run the jobs and
to run some test steps.


tests/kuttl/tests/change_keystone_config/00-deploy_keystone.yaml
```diff
@@ -0,0 +1 @@
+../../common/keystoneapi_deploy.yaml
```
stuggi, line 1: should there always be cleanup be triggered as 00 to make sure we start with fresh sample deployment?

stuggi (commented):
we should probably rename the install_yaml `keystone_deploy_validate` target to something else, like keystone_kuttl as we no longer validate the deployment created with keystone_deploy ?

tests/kuttl/tests/change_keystone_config/00-deploy_keystone.yaml
cescgina: Done

tests/kuttl/common/keystoneapi_deploy.yaml
```diff
@@ -0,0 +1 @@
+../../../config/samples/keystone_v1beta1_keystoneapi.yaml
```
marios, line 1: interesting... so we don't need to issue any oc commands... like kuttl will just apply the crd we have defined there? https://github.com/openstack-k8s-operators/keystone-operator/blob/b9019b4533f4049d41677329134d6dddb0ab0a63/config/samples/keystone_v1beta1_keystoneapi.yaml#L1

cescgina: Exactly, if we pass a yaml file with kubernetes resources, kuttl will apply them to the cluster

abays:
This PR should also update `kuttl-test.yaml`'s comment [1] on how to run the KUTTL tests.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/0c0446d0c30868b0698a0e31dffc02de6f536e16/kuttl-test.yaml#L4

cescgina:
> This PR should also update `kuttl-test.yaml`'s comment [1] on how to run the KUTTL tests.
> 
> [1]
> 
> https://github.com/openstack-k8s-operators/keystone-operator/blob/0c0446d0c30868b0698a0e31dffc02de6f536e16/kuttl-test.yaml#L2-L5

I've updated the comment since it was outdated and the target name changed. To clarify, however, we will still run the kuttl tests from install_yamls (at least in CI).

abays:
> > This PR should also update `kuttl-test.yaml`'s comment [1] on how to run the KUTTL tests.
> > [1]
> > https://github.com/openstack-k8s-operators/keystone-operator/blob/0c0446d0c30868b0698a0e31dffc02de6f536e16/kuttl-test.yaml#L2-L5
> 
> I've updated the comment since it was outdated and the target name changed. To clarify, however, we will still run the kuttl tests from install_yamls (at least in CI).

Ah, thank you for clarifying.

abays (approved):
/lgtm

PR #227, opened 2023-04-05
Remove fmt dependency from make tidy

gibizer:
The fmt dependency cannot be run if the go.mod file needs an update:
```
 $ make tidy
 go fmt ./...
 go: updates to go.mod needed; to update it:
 	go mod tidy
 make: *** [Makefile:103: fmt] Error 1
```


abays (approved):
/lgtm

PR #231, opened 2023-04-13
Add native interface to customize policy

kajinamit:
This introduces the new native interface so that users can inject customized policy rules. This interface depends on the capability of oslo.policy which loads separate files from policy.d directory. This is much flexible than using the base policy.yaml file because it allows us to use the default policy rules installed by rpm (eg. octavia) and override only specific rules.


bshephar (commented):
Maybe we should have a sample file with an example policy.yaml string that leverages this new interface?

PR #234, opened 2023-04-21
Disable log file

kajinamit:
This change disables usage of log file and makes all log outputs sent to stderr. The file is created inside pod and can fill up container file system.


abays (approved):
/lgtm

PR #240, opened 2023-05-03
Change kuttl tests to run in arbitrary namespace

cescgina:
Kuttl tests could originally only be run in the openstack namespace.
This change allows them to run in any namespace, by default set to
keystone-kuttl-tests, to potentially avoid clashes with dev
environments. The namespace is still set at runtime and can't be managed
by kuttl since the resources keystone-operator depends on (i.e. mariadb)
must be deployed in the same namespace.

To run this tests, the following PR from install_yamls should be used:
https://github.com/openstack-k8s-operators/install_yamls/pull/240


abays (approved):
Tested and passed

/lgtm

PR #246, opened 2023-05-18
Assert CronJob resource in kuttl tests

kajinamit:
Current keystone-operator deploys a CronJob resource to purge expired trusts. This adds a simple assertion to make sure that resource is created.


tests/kuttl/tests/keystone_scale/03-assert.yaml
```diff
@@ -24,3 +24,11 @@ spec:
...
 status:
   availableReplicas: 3
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: keystone-cron
+spec:
+  schedule: 1 * * * *
+  suspend: false
```
abays, line 34: nit: I know it's obvious from just reading the file, but could you put the `CronJob` in the list comment at the top?

kajinamit: Yes. That makes sense. The line was added in the latest PR.

abays (approved):
/lgtm

PR #250, opened 2023-05-23
Reset Ready condition when update failed

kajinamit:
This ensures the Ready Condition is reset when CR becomes once ready but something fails during reconfiguration.

### DIFF
--- a/api/v1beta1/keystoneapi_types.go
+++ b/api/v1beta1/keystoneapi_types.go
@@ -269,10 +269,9 @@ func (instance KeystoneAPI) GetEndpoint(endpointType endpoint.Endpoint) (string,
 	return "", fmt.Errorf("%s endpoint not found", string(endpointType))
 }
 
-// IsReady - returns true if service is ready to server requests
+// IsReady - returns true if KeystoneAPI is reconciled successfully
 func (instance KeystoneAPI) IsReady() bool {
-	return instance.Status.Conditions.IsTrue(condition.ExposeServiceReadyCondition) &&
-		instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition)
+	return instance.Status.Conditions.IsTrue(condition.ReadyCondition)
 }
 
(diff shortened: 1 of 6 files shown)


api/v1beta1/keystoneapi_types.go
```diff
@@ -269,10 +269,10 @@ func (instance KeystoneAPI) GetEndpoint(endpointType endpoint.Endpoint) (string,
 	return "", fmt.Errorf("%s endpoint not found", string(endpointType))
 }
 
-// IsReady - returns true if service is ready to server requests
+// IsReady - returns true if KeystoneAPI is reconciled successfully
 func (instance KeystoneAPI) IsReady() bool {
-	return instance.Status.Conditions.IsTrue(condition.ExposeServiceReadyCondition) &&
-		instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition)
+	readyCond := instance.Status.Conditions.Get(condition.ReadyCondition)
+	return readyCond != nil && readyCond.Status == corev1.ConditionTrue
```
abays, line 275: Could we just use `instance.Status.Conditions.IsTrue(condition.ReadyCondition)` here?  It seems like handles the `nil` case and then the boolean comparison [1].

[1] https://github.com/openstack-k8s-operators/lib-common/blob/d41eddbb4da64c307f3b510358f09a54a909b84d/modules/common/condition/funcs.go#L135-L142

kajinamit: That is definitely better. Thanks for pointing that !

bshephar (approved):
lgtm

abays (approved):
/lgtm

This is probably something we want to replicate elsewhere too

kajinamit:
> /lgtm
> 
> This is probably something we want to replicate elsewhere too

Yes. The current plan is to replicate this to all operators. This also helps us remove ServiceID (keystone service id) which is now used in IsReady method but can be replaced by checking ReadyCondition.

PR #255, opened 2023-05-29
Prohibit changing name of admin role

kajinamit:
The policy rules implemented in each service expect the admin role is named "admin", and the policy enforcement results in unexpected denials if we use a different role name.


abays (approved):
/lgtm

PR #258, opened 2023-06-08
Update renovate.json

kajinamit:
The current content is different from the other operators. This updates it to have the consistent setting across operators.


stuggi (approved):
/lgtm

PR #259, opened 2023-06-08
CSV: Set operator-type: non-standalone

dprince:
This should hide the keystone package/bundle in the OpenShift console as we only want the main OpenStack operator to show up there.


abays (approved):
/lgtm

PR #261, opened 2023-06-23
Allow configuring timezone

kajinamit:
This introduces the TimeZone spec to allow users to customize timezone used by pods, so that logs are recorded with local time.

### DIFF
--- a/api/bases/keystone.openstack.org_keystoneapis.yaml
+++ b/api/bases/keystone.openstack.org_keystoneapis.yaml
@@ -245,6 +245,10 @@ spec:
                 description: Secret containing OpenStack password information for
                   keystone KeystoneDatabasePassword, AdminPassword
                 type: string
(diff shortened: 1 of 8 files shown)


bshephar (commented):
This sounds good. I think users will definitely want the ability to do this. 

If we extrapolate this implementation out though, we would need to provide a TZ for each and every service. It would be nice if we ultimately had a way to set the TZ once at the top level of OpenStackControlPlane and have it apply to all of the operators.

kajinamit:
> so that logs are recorded with local time.
This might not be completely true. Log metadata k8s/openshift adds are all in UTC so adding tz only affects timestamps in log messages.

kajinamit:
We need to discuss whether this is really a required feature, based on the fact this only affects log timestamp. I'll close this atm to clear up review lists.

PR #262, opened 2023-06-26
Prohibit updating spec.databaseInstance

kajinamit:
Updating spec.databaseInstance does not migrate existing data to a new instance and can break resource access unexpectedly. This prohibits such update.

### DIFF
--- a/api/v1beta1/keystoneapi_webhook.go
+++ b/api/v1beta1/keystoneapi_webhook.go
@@ -23,8 +23,14 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


abays:
I wonder if the new "immutable" tag available in Kubernetes 1.25+ would be cleaner here.  An example can be found here [1].  While in this example they're making the whole CRD immutable, I think you can apply the `kubebuilder` annotation to just one particular field.

[1] https://blog.rewanthtammana.com/kubernetes-crd-validation-with-cel-and-kubebuilder-marker-comments#heading-crd-validation-in-kubernetes-125

kajinamit:
Yeah using that validation feature can be much simpler. The only concern is that the feautre is still in beta state since it was promoted at 1.25, afaik.

PR #263, opened 2023-06-26
Ensure KeystoneAPI is ready in kuttl tests

kajinamit:
This ensures the Ready condition of KeystoneAPI becomes ready in every deployment step. Note that the condition should be false in case no pod is available so it is intentionally asserted to be false in case keystone service is scaled down to 0.

### DIFF
--- a/tests/kuttl/common/assert_sample_deployment.yaml
+++ b/tests/kuttl/common/assert_sample_deployment.yaml
@@ -40,6 +40,9 @@ spec:
 status:
   databaseHostname: openstack
   readyCount: 1
+  conditions:
+  - status: "True"
+    type: Ready
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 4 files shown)


tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -40,6 +40,9 @@ spec:
 status:
   databaseHostname: openstack
   readyCount: 1
+  conditions:
+  - status: "True"
+    type: Ready
```
abays, line 45: This won't work with KUTTL because it requires YAML lists to be in an exact order and have the exact number of items:

```
case.go:231: resource KeystoneAPI:keystone-kuttl-tests/keystone: .status.conditions: slice length mismatch: 1 != 13
```

kajinamit: Hmm. OK so we need to query a specific key instead. I'll postpone this now but will look into any better solutions when I get time.

abays: One way to do this is to check for the associated `KeystoneAPI` service pod that has this label and this status:

```
metadata:
...
  labels:
    service: keystone
...
status:
  ...
  phase: Running
  ...
```

Or we could update the `Deployment` created for the `KeystoneAPI` to give it that label as well (it currently lacks it), and then check its `Status.ReadyReplicas`.

PR #264, opened 2023-06-29
Remove redundant default of PasswordSelectors

kajinamit:
Currently we define defaults in two layers but this is redundant.

### DIFF
--- a/api/bases/keystone.openstack.org_keystoneapis.yaml
+++ b/api/bases/keystone.openstack.org_keystoneapis.yaml
@@ -159,9 +159,6 @@ spec:
                   this service
                 type: object
               passwordSelectors:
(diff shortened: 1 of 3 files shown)


kajinamit:
The core problem seems to be that when passwordSelector is not given then it takes an empty object and the value from that empty object is not omitted because of absence of omitempty. Because we are not allowing default with omitempty enabled, it might be impossible to achieve this. I didn't notice this when I was implementing this in heat-operator, because we used very old version of operator-lint which allowed that usage.

Wondering if using an empty dict as a default works. If that does not work then I might just abandon this.

kajinamit:
I tried a few more patterns but none of these worked. So I'll abandon this.

PR #265, opened 2023-06-30
Do not delete KeystoneService with associated KeystoneEndpoint

kajinamit:
The KeystoneEndpoint CR is dependent on the KeystoneService CR, because keystone endpoints requires that the target service is created in advance.

This adds the finalizer to avoid removing KeystoneService before all associated KeystoneEndpoint instances are deleted.


kajinamit:

The issue with Galera assertion is supposed to be fixed now.

abays (approved):
/lgtm

PR #269, opened 2023-07-04
Allow customize route via spec.Override.Route

stuggi:
Allows to customize the service route vie the `spec.Override.Route`. This allows e.g. to add custom labels, configure the route via annotations as in [1], or set TLS parameters.

~~~
kind: KeystoneAPI
metadata:
  name: keystone
  namespace: openstack
spec:
  ...
  override:
    routeOverride:
      metadata:
        annotations:
          haproxy.router.openshift.io/timeout: "60"
        labels:
          mylabel: boo
~~~

[1] https://docs.openshift.com/container-platform/4.13/networking/routes/route-configuration.html#nw-route-specific-annotations_route-configuration

Jira: [OSP-21715](https://issues.redhat.com//browse/OSP-21715)
Jira: [OSP-26299](https://issues.redhat.com//browse/OSP-26299)


api/v1beta1/keystoneapi_types.go
```diff
@@ -147,6 +148,16 @@ type KeystoneAPISpec struct {
...
+
+	// +kubebuilder:validation:Optional
+	// Override, provides the ability to override the generated manifest of several child resources.
+	Override KeystoneAPIOverrideSpec `json:"override,omitempty"`
+}
+
+// KeystoneAPIOverrideSpec to override the generated manifest of several child resources.
+type KeystoneAPIOverrideSpec struct {
+	// Override configuration for the Route created to serve traffic to the service
+	Route *route.OverrideSpec `json:"route,omitempty"`
```
gibizer, line 160: So lib-common does a strategicpatch.StrategicMergePatch between the original route spec an this override. I think we still need to define for the user what behavior that means. 
My guesses / questions:
* the list of labels and annotations are merged
* the host, subdomain and path fields are overridden in the original spec if defined here
* Does the `to` field which is a struct RouteTargetReference overridden field by field or as a single entity. I.e can I override the to.weight independently without defining the to.name  or I need to define both in the override?
* AlternateBackends is a list. Does the items in that list merged with the items in the override?
* ... you probably interpolate my questions about the behavior of the rest of the fields.

stuggi: > So lib-common does a strategicpatch.StrategicMergePatch between the original route spec an this override. I think we still need to define for the user what behavior that means. My guesses / questions:
> 
> * the list of labels and annotations are merged

correct, new labels/annotations get merged with the ones created by the operator. If the user provides the same annotation/label as the operator creates, the one from the user will overwrite the one from the operator.

> * the host, subdomain and path fields are overridden in the original spec if defined here
correct

> * Does the `to` field which is a struct RouteTargetReference overridden field by field or as a single entity. I.e can I override the to.weight independently without defining the to.name  or I need to define both in the override?

you have to override the full struct since those are required values, otherwise we have to create a copy of them as well and make them all optional.

> * AlternateBackends is a list. Does the items in that list merged with the items in the override?

this depends on the list patchStrategy. https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch . so this is different from list to list. if `patchStrategy:"merge` is set they get merged, otherwise they get replaced.

> * ... you probably interpolate my questions about the behavior of the rest of the fields.

in general 
* required parameters of sub structs have to be named, otherwise we have to created copies of them and make copies and make all parameters optional. 
* I tried to keep it simple for this initial approach and just make a copy of the top level and make required parameter optional
* for lists as mentioned above it depends on the patchStrategy . an  example of those can be found at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podspec-v1-core where you see Container has merge as strategy set. when you scroll through the podSpec, e.g. readinessGates does not has merge set.

I'll try to add more tests to lib-common which should cover those scenarios

stuggi: added more tests in lib-common

gibizer: Thank you for the explanation and the extra test coverage. Can we document somehow the behavior you described above  in the doc of this field so that the end user will know what to expect?

stuggi: yes that makes sense. I'll do that

stuggi: updated the lib-common PR to describe the behavior

gibizer (commented):
This looks OK to me now. I will go and review the lib-common dependency.

rabi:
@stuggi Any plan for rebasing this and the lib-common PR soon? Looking at the frequent 504 gateway timeouts in jobs[1], I suspect we need to increase `haproxy.router.openshift.io/timeout` from the default 30s(?).

[1] https://logserver.rdoproject.org/10/410/740a298105dfd2f689f5ce2e36d22c369ec2ae33/github-check/openstack-operator-crc-podified-edpm-baremetal/d17206e/job-output.txt

gibizer:
After talking to @stuggi this might not needed in the service operators if we move the route creation to the openstack-operator level instead. That seems to be an idea worth to explore.

stuggi:
addressed by https://github.com/openstack-k8s-operators/openstack-operator/pull/416

PR #271, opened 2023-07-06
Watch for Memcached

abays:
Since we now pass a `Memcached` instance name to the `KeystoneAPI`, its controller should watch for changes to `Memcached`s.


kajinamit:
Thanks for catching this ! Yes. we need this (I have to add the same to horizon-operator).

PR #273, opened 2023-07-11
Control deletion of service dependencies

abays:
We have been seeing an error in `OpenStackControlPlane` deletion where Keystone and MariaDB resources linger without being fully cleaned-up.  This is due to Keystone now using Memcached.  When the `OpenStackControlPlane` is deleted, what happens is that `Memcached` and its pod are deleted before the deletion logic in Keystone can complete.  This results in the reconcile loops of `KeystoneEndpoint` and `KeystoneService` hitting server errors when calling the admin client to remove their entries on the OpenStack side of things  (because the cache disappeared and the Keystone API can't find it, so it throws a 500 error).  Then, due to these errors, the OpenShift side does not ever finish cleaning up the associated resources.

This patch introduces two things:

1. `KeystoneAPI` now adds a finalizer to its `Memcached` dependency resource to prevent it from being deleted while the `KeystoneAPI` is still in service.
2. The `KeystoneAPI` does not remove its finalizer from `Memcached` _nor_ MariaDB/Galera _nor_ itself until the `KeystoneEndpoint` and `KeystoneService` CRs have themselves removed their finalizers from the `KeystoneAPI` CR.

I worry that this new logic will introduce hangs elsewhere, but I have not seen anything so far in my testing.  Those tests have by no means been exhaustive, however.  The purpose of this PR is, at the very least, to get a conversation going in regards to this issue even if we don't go with this solution.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -101,7 +101,8 @@ type KeystoneAPIReconciler struct {
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=route.openshift.io,resources=routes,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbdatabases,verbs=get;list;watch;create;update;patch;delete;
-// +kubebuilder:rbac:groups=memcached.openstack.org,resources=memcacheds,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=memcached.openstack.org,resources=memcacheds,verbs=get;list;watch;update
+// +kubebuilder:rbac:groups=memcached.openstack.org,resources=memcacheds/finalizers,verbs=update
 // +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
 
 // service account, role, rolebinding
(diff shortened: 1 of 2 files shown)


rabi (approved):
Yeah I did notice the same thing yesterday and was planning to to add finalizer on memcached. Shared memcached instance should not be deleted before the services using it and every service using it ideally should have a finalizer on it. Though I think services should be able to ignore memcached not found errors during delete reconciliation. But I guess they watch for memcached atm and get into that reconciliation loop  which makes it difficult.

rabi:
I guess we've to do it in all services where the memcached service is used like db[1].

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/controllers/glance_controller.go#L212

kajinamit:
I agree we need this, but this would cause a problem in case a user updates the `memcachedInstance` spec, because we do not remove the finalizer from the old memcached instance. This is a problem similar to the one I tried to point out when I was working on the webhook validation to prohibit change in databaseInstance.

We may probably record the old memcachedInstance name in status, or query all memcacheds and remove unnecessary finalizers. However the 2nd approach is costly so probably the 1st one is better. Attempting it in https://github.com/openstack-k8s-operators/keystone-operator/pull/274

gibizer:
I see I late to the party as this get merged already.

+1 on adding the keytoneapi finalizer to the memcached it is using
+1 on not removing the keystoneapi finalizer from memcached while there are keystoneservice and keystoneendpoint instances exists. 

The general logic of delete of a CR is
# Do nothing if there are foreign finalizers on the CR. This means somebody is still using this CR.
# If only the CR's own finalizer left on the CR then start cleaning up the resources this CR uses. This includes (but not limited to) removing the CR's finalizer from other resources.
# After every resource this CR uses are cleaned up, remove its own finalizer from this CR.

controllers/keystoneapi_controller.go
```diff
@@ -266,6 +267,39 @@ func (r *KeystoneAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		// there is more Keystone processing that needs to finish before we can
+		// allow our DB and Memcached dependencies to be potentially deleted
+		// themselves
+		if finalizer != helper.GetFinalizer() {
+			return ctrl.Result{}, nil
+		}
+	}
+
+	// Remove our finalizer from Memcached
+	memcached, err := r.getKeystoneMemcached(ctx, helper, instance)
```
stuggi, line 288: getKeystoneMemcached could/should be a generic client func getMemcachedByName  in the infra-operator memcached api, or a client module in there, like we have for database in lib-common. we discussed in the past to move the database module from lib-common to mariadb-operator as it is tightly coupled with it.

controllers/keystoneapi_controller.go
```diff
@@ -266,6 +267,39 @@ func (r *KeystoneAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	for _, finalizer := range instance.Finalizers {
+		// If this finalizer is not our KeystoneAPI finalizer, then it is either
+		// a KeystoneService or KeystoneEndpointer finalizer, which indicates that
+		// there is more Keystone processing that needs to finish before we can
+		// allow our DB and Memcached dependencies to be potentially deleted
+		// themselves
+		if finalizer != helper.GetFinalizer() {
+			return ctrl.Result{}, nil
+		}
+	}
```
stuggi, line 285: should this be a simple lib-common func and a generic check in the all the operators being called right at the top of delete?

controllers/keystoneapi_controller.go
```diff
@@ -266,6 +267,39 @@ func (r *KeystoneAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	if !k8s_errors.IsNotFound(err) && memcached != nil {
+		if controllerutil.RemoveFinalizer(memcached, helper.GetFinalizer()) {
+			err := r.Update(ctx, memcached)
+
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	}
```
stuggi, line 301: should the also be a func in the memcached module, like `db.DeleteFinalizer`

PR #274, opened 2023-07-12
Remove finalizer from previous memcached instance

kajinamit:
This ensures finalizer is removed from the old memcached instance in case a user updates the memcachedInstance spec. Otherwise finalizer is not removed and users can't delete the instance once used by KeystoneAPI.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -194,6 +194,7 @@ func (r *KeystoneAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		// Register overall status immediately to have an early feedback e.g. in the cli
 		return ctrl.Result{}, nil
 	}
+
 	if instance.Status.Hash == nil {
 		instance.Status.Hash = map[string]string{}
 	}
@@ -290,14 +291,18 @@ func (r *KeystoneAPIReconciler) reconcileDelete(ctx context.Context, instance *k
 		return ctrl.Result{}, err
 	}
 
-	if !k8s_errors.IsNotFound(err) && memcached != nil {
+	if !k8s_errors.IsNotFound(err) {
 		if controllerutil.RemoveFinalizer(memcached, helper.GetFinalizer()) {
 			err := r.Update(ctx, memcached)
-
 			if err != nil {
 				return ctrl.Result{}, err
 			}
 		}
+	}
+
+	err = r.removeFinalizerFromOldKeystoneMemcacheds(ctx, helper, instance)
+	if err != nil {
+		return ctrl.Result{}, err
 	}
 
 	// remove db finalizer before the keystone one
@@ -857,6 +862,18 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+	// create OpenStackClient config - end
+
+	err = r.removeFinalizerFromOldKeystoneMemcacheds(ctx, helper, instance)
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			keystonev1.KeystoneMemcachedReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			keystonev1.KeystoneMemcachedReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
 
 	l.Info("Reconciled Service successfully")
 	return ctrl.Result{}, nil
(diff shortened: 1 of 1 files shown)


api/v1beta1/keystoneapi_types.go
```diff
@@ -231,6 +231,9 @@ type KeystoneAPIStatus struct {
 
 	// NetworkAttachments status of the deployment pods
 	NetworkAttachments map[string][]string `json:"networkAttachments,omitempty"`
+
+	// Memcached instance name.
+	MemcachedInstance string `json:"memcachedInstance,omitempty"`
```
rabi, line 236: Let's understand why we would need this  and what're the use-cases?  I don't expect this to be changed in production deployments.

kajinamit: This is the field in status, not the one in spec. Because we have to find out the instance previously used, so I decided to record it in the spec.

Talking about the spec.MemcachedInstance field, that follows the common approach to handle shared services such as db or mq. I agree we 'd avoid having different approach specific to memcached.

rabi: Ack, I did not know that we've already added that in spec. I don't think we should all these assuming someone would flip between memached instances.

kajinamit: If we don't assume people would not flip between memcached instance, then we should prohibit that operation (by adding validation at webhook). I can add the validation, but that would also trigger the discussion about the potential migration scenario (similarly to db, though we less likely to see the use case specific to memcached).

Because the fix is light, I'd prefer keeping the capability now.

controllers/keystoneapi_controller.go
```diff
@@ -646,6 +657,18 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
...
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			keystonev1.KeystoneMemcachedReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			keystonev1.KeystoneMemcachedReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+	instance.Status.MemcachedInstance = instance.Spec.MemcachedInstance
+
```
gibizer, line 671: In the following unlikely sequence of events I think this code will leak memcached finalizers:
1. Spec.MemcachedInstance=old Status.MemcachedInstance=old and everything is Ready and stable
2. user set Spec.MemcachedInstance=new
3. Reconcile#1 start with Spec.MemcachedInstance=new Status.MemcachedInstance=old
4. Reconcile#1 adds finalizer to MemcachedInstance=new L646
5. Reconcile#1 removeFinalizerFromOldKeystoneMemcached() reads MemcachedInstance=old from etcd 
6. user removes the finalizer and deletes MemcachedInstance=old manually
7. Reconcile#1 removeFinalizerFromOldKeystoneMemcached() tries to update MemcachedInstance=old to remove the finalizer and gets NotFound
8. NotFound propagated to L668 and Reconcile#1 exists ( Spec.MemcachedInstance=old Status.MemcachedInstance=new is the persisted state) 
9. user updates Spec.MemcachedInstance=new2
10. Reconcile#2 starts and reads instance from etcd with Spec.MemcachedInstance=new2 Status.MemcachedInstance=old

At this point MemcachedInstance=new has our finalizer but it is not recorded in Status.MemcachedInstance so that finalizer is never removed by this operator.
This needs two precisely timed user interaction at 6) and 9) so this is an unlikely sequence.

gibizer: There is another variant of the above. If at 6) the something updates MemcachedInstance=old and therefore bumps the resource generation of it. Then at 7) we get Conflict and if the rest of the sequence is kept as is then we still end up leaking the finalizer on MemcachedInstance=new. While the NotFound case can be easily fixed by not treating NotFound as an error at 7). The Conflict scenario cannot be similarly fixed.

abays:
Thanks for doing this.  I had it in mind as a follow-up when I added #273 yesterday, but I was thinking of doing it via webhooks.  I think this works fine though.

gibizer (commented):
I feel we are growing complexity for something that is not really a high priority user demand. It might be simpler and safer to forbid changing the memcached instance for now.

controllers/keystoneapi_controller.go
```diff
@@ -1096,3 +1119,35 @@ func (r *KeystoneAPIReconciler) getKeystoneMemcached(
...
+		},
+		memcached)
+
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
+
+	if !k8s_errors.IsNotFound(err) {
+		if controllerutil.RemoveFinalizer(memcached, h.GetFinalizer()) {
+			return r.Update(ctx, memcached)
```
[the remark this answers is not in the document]
abays: Yes, catching `NotFound` here and ignoring it makes sense to me too

controllers/keystoneapi_controller.go
gibizer: Ohh. Good question. If we remove the finalizer from the old memcached first, then we have to make sure that at that point the keystone openstack service is already reconfigured to use the new memcached instance as the old can disappear. I haven't traced this case through yet...

controllers/keystoneapi_controller.go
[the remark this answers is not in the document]
kajinamit: Fixed in the current version

controllers/keystoneapi_controller.go
kajinamit: I eventually decided to go through all Memcached instances... that would be more robust and would save us from considering all corner cases.

Also, it allows us to do cleanup after all resources are reconciled, to ensure the memcached instance used by the current config file is kept before config files are updated and services are restarted.

abays:
> I feel we are growing complexity for something that is not really a high priority user demand. It might be simpler and safer to forbid changing the memcached instance for now.

I'm okay with blocking the Memcached swap via webhooks.  Is everyone else here on board with that?  If not, should we discuss this in the sync tomorrow?

rabi:
> I'm okay with blocking the Memcached swap via webhooks. Is everyone else here on board with that? If not, should we discuss this in the sync tomorrow?

Not sure if I would be able to join the sync today.  +1 from me, I would prefer we keep things simple in design (if that means not supporting some corner cases directly).

kajinamit:
I'm ok with prohibiting update of memcached, though I'm not too sure then if we want to really support splitting database with using a single memcached, which sounds like inconsistent.

PR #275, opened 2023-07-14
Remove duplicate jobs from github workflow

kajinamit:
We now run pre-commit job in prow which covers most of go-lang tasks.


abays (approved):
/lgtm

PR #277, opened 2023-07-18
Replace KeystoneMemcachedReadyCondition by common condition

kajinamit:
We have added the common MemcachedReadyCondition to lib-common to reduce duplicate implementations.

Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/300


abays (approved):
/lgtm

PR #278, opened 2023-07-18
Add OpenStackClient CR

kajinamit:
This migrates OpenStackClient CR from infra-operator to resolve circular dependencies between infra-operator and keystone-operator.

Depends-on: https://github.com/openstack-k8s-operators/infra-operator/pull/90


abays:
This won't be able to pass CI, so we might have to force-merge it:

```
keystone-operator-index/openstack-operators/alpha/keystone-operator.v0.0.1
      and @existing/openstack-operators//infra-operator.v0.0.1 provide OpenStackClient
      (client.openstack.org/v1beta1)'
```

abays:
> This won't be able to pass CI, so we might have to force-merge it:
> 
> ```
> keystone-operator-index/openstack-operators/alpha/keystone-operator.v0.0.1
>       and @existing/openstack-operators//infra-operator.v0.0.1 provide OpenStackClient
>       (client.openstack.org/v1beta1)'
> ```

Maybe it will now that we merged https://github.com/openstack-k8s-operators/openstack-operator/pull/397

jistr:
FYI the Adoption job now depends on having the client pod deployed, so the CI is broken until we re-introduce it. Should the CR move take a long-ish time to complete, we should probably revert to an older way of using the CLI in the Adoption test suite.

jistr:
That Adoption refactoring is actually embedded in a patch with other refactorings, so it may be actually easier to revert the change in openstack-operator for now, so i proposed it: https://github.com/openstack-k8s-operators/openstack-operator/pull/398

abays:
> That Adoption refactoring is actually embedded in a patch with other refactorings, so it may be actually easier to revert the change in openstack-operator for now, so i proposed it: [openstack-k8s-operators/openstack-operator#398](https://github.com/openstack-k8s-operators/openstack-operator/pull/398)

This particular Keystone patch won't pass CI until we also patch the OpenStack operator to include the new Infra operator with the `OpenStackClient` CRD removed.  Otherwise both this new Keystone operator and the old Infra operator will collide due to the presence of the `OpenStackClient` CRD in both, which will prevent the OpenStack operator from installing and thus block CI for this patch.  So it seems we're in a situation where Adoption needs the old Infra to be used by OpenStack operator, but this new patch needs the new Infra.  We'll have to coordinate something.

jistr:
Ack. Yatin mentioned:

```
oc apply -f https://github.com/openstack-k8s-operators/infra-operator/raw/82e0480408a53cb5c7e91cdc138279884d5ffdcc/config/samples/client_v1beta1_openstackclient.yaml

oc rsh test
```

Can we rely on that working during the whole transition period?

Alternatively we could indeed do the selective revert of the Adoption test suite to use local openstack CLI again instead of the podified one.

kajinamit:
> Ack. Yatin mentioned:
> 
> ```
> oc apply -f https://github.com/openstack-k8s-operators/infra-operator/raw/82e0480408a53cb5c7e91cdc138279884d5ffdcc/config/samples/client_v1beta1_openstackclient.yaml
> 
> oc rsh test
> ```
> 
> Can we rely on that working during the whole transition period?
> 
> Alternatively we could indeed do the selective revert of the Adoption test suite to use local openstack CLI again instead of the podified one.

No OpenStackClient CR will be removed from infra-operator so it would no longer work at that point. Later it would be supported by keystone-operator and CR would be available again, but the sample file would be also moved to keystone-operator so the file link should be updated at that time.

gibizer (approved):
looks sane

gibizer (approved):

api/client/v1beta1/conditions.go
```diff
@@ -0,0 +1,53 @@
...
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+)
+
+// OpenStackClient Condition Types used by API objects.
```
kajinamit, line 22: the previous version contained `OpenStackControlPlane` Condition and Messages but these are not really used (I suspect that was unintentionally brought to infra-operator when we initially moved client cr from openstack-operator).

The latest version have these all removed. I didn't catch any problems in local sanity checks.

SeanMooney:
i feel like this is the wrong approach.

i think this should be move to the opesntack operator not the keystone operator

PROJECT
```diff
@@ -38,4 +38,17 @@ resources:
   kind: KeystoneEndpoint
   path: github.com/openstack-k8s-operators/keystone-operator/api/v1beta1
   version: v1beta1
```
dprince, line 40: since we now have 'multiple groups' in the same operator should we mark it as multigroup: true? NOTE: this would fundamentally change the structure of the api's directory to accommodate multiple groups as well.

dprince (changes requested):

dprince:
one potential problem with this is that we are essentially treating keystone-operator as a 'multigroup: true' operator but is isn't structured as such. And isn't marked that way in PROJECT https://github.com/openstack-k8s-operators/infra-operator/blob/main/PROJECT#L4 
Only the infra-operator and openstack-operator are multigroup I think as far as openstack-k8s-operators goes.

kajinamit:
/hold

based on the feedback from Dan, I think it's better to put this to openstack-operator now, instead of converting the structure of keystone-operator now. We might hear some feedback especially related to dev workflow but we can revisit the requirement(there were some questions raised earlier if we really need this automated by operator or we can replace this by a standard resource definition yaml) at that time.

PR #281, opened 2023-07-20
Update to use shared renovate config

dprince:


renovate.json
```diff
@@ -1,47 +1,5 @@
...
-    },
-    {
-      "groupName": "sigs.k8s.io/controller-runtime",
-      "matchPackagePatterns": ["^sigs.k8s.io/controller-runtime"],
-      "allowedVersions": "< 0.15.0",
-      "enabled": true
-    }
-  ],
-  "postUpgradeTasks": {
-    "commands": ["make gowork", "make tidy", "make manifests generate"],
```
gibizer, line 43: It seems the common config regresses this as it only runs make tidy https://github.com/openstack-k8s-operators/renovate-config/blob/79f5b0ba87339a43b8c7bb9f9391aa3f7152fab1/default.json#L47

dprince: okay, so I guess we can drop postUpgrade from common and then just have it in each operator. Probably safer that way as it depends on the Makefile too right

dprince: https://github.com/openstack-k8s-operators/renovate-config/pull/1

stuggi: +1 sounds good to me

gibizer: I'm OK with that as I can imagine some repository does not have manifests and generate targets.

renovate.json
```diff
@@ -1,43 +1,6 @@
 {
-"extends": [
-    "config:base"
-  ],
-  "dependencyDashboard": true,
-  "logFileLevel": "trace",
-  "enabledManagers": ["gomod"],
-  "postUpdateOptions": ["gomodTidy"],
-  "constraints": {
-    "go": "1.19"
```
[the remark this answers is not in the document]
dprince: Good call. Lets keep go in this file too.

stuggi:
+1 from me

gibizer (commented):
looks good to me too

stuggi (approved):
/lgtm

PR #283, opened 2023-07-21
Use pre-created osp-secret in kuttl tests

kajinamit:
Now install_yamls creates the global osp-secret in the namespace where kuttl tests run. This replaces the specific secret created for tests to and make that global secret used consistently.

This also removes the step to clean up existing KeystoneAPI CR, because kuttl targets now contains the required cleanup steps.


kajinamit:

503 while downloading telemetry-operator image

abays (approved):
Tested locally

/lgtm

PR #287, opened 2023-07-24
Remove config/manager/kustomization.yaml from .gitignore

kajinamit:
This file is required to build a bundle image. The file is actually maintained in the repository.


kajinamit:
This is intentional to avoid pulling changes by `make bundle`.

PR #289, opened 2023-07-28
Update keystoneapi to use service override

stuggi:
Removes creation of routes for keystoneapi as those get done in the openstack-operator. Via service overrides the openstack-operator customize the services for the endpoints to the desired state specified in the ctlplane CR using externalEndpoints or the service override in the keystone template.

Jira: [OSP-26690](https://issues.redhat.com//browse/OSP-26690)

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/313


controllers/keystoneapi_controller.go
```diff
@@ -452,59 +451,89 @@ func (r *KeystoneAPIReconciler) reconcileInit(
...
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				condition.ExposeServiceReadyRunningMessage))
+			return ctrlResult, nil
+		}
+		// create service - end
+
+		// TODO: TLS, pass in https as protocol, create TLS cert
+		apiEndpoints[string(endpointType)], err = svc.GetAPIEndpoint(
+			svcOverride, data.Protocol, data.Path)
```
gibizer, line 526: Does TLS needs work in every service operator? I thought one major reason to do all this work is to be able to implement TLS in a single place.

controllers/keystoneapi_controller.go
```diff
@@ -452,59 +451,89 @@ func (r *KeystoneAPIReconciler) reconcileInit(
...
 
-	for _, metallbcfg := range instance.Spec.ExternalEndpoints {
-		portCfg := keystonePorts[metallbcfg.Endpoint]
-		portCfg.MetalLB = &endpoint.MetalLBData{
-			IPAddressPool:   metallbcfg.IPAddressPool,
-			SharedIP:        metallbcfg.SharedIP,
-			SharedIPKey:     metallbcfg.SharedIPKey,
-			LoadBalancerIPs: metallbcfg.LoadBalancerIPs,
+	for endpointType, data := range keystoneEndpoints {
+		endpointName := keystone.ServiceName + "-" + string(endpointType)
```
gibizer, line 467: This feels duplicated information between here and https://github.com/openstack-k8s-operators/openstack-operator/pull/416/files#diff-dc49c70e72e8b711bfefebe68ede374400ee1c2c0fe24c4a25a1a6c352a1a489R128

api/v1beta1/keystoneapi_types.go
```diff
@@ -150,37 +151,14 @@ type KeystoneAPISpec struct {
...
-	// SharedIP is true, but no SharedIPKey specified.
-	SharedIPKey string `json:"sharedIPKey"`
-
-	// +kubebuilder:validation:Optional
-	// LoadBalancerIPs, request given IPs from the pool if available. Using a list to allow dual stack (IPv4/IPv6) support
-	LoadBalancerIPs []string `json:"loadBalancerIPs,omitempty"`
+// APIOverrideSpec to override the generated manifest of several child resources.
+type APIOverrideSpec struct {
+	// Override configuration for the Service created to serve traffic to the cluster.
+	Service []service.OverrideSpec `json:"service,omitempty"`
```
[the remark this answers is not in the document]
stuggi: yes, correct, it is an override per endpoint type the operator supports. I was considering to switch to a single service (with multiport if the OSP service requires), but Sean mentioned in a discussion we had that it might be useful/required to use different LoadBalancer service types per endpoint. I kept it to be a list to align with what we have for externalEndpoints if that makes sense, e.g. https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml#L25

controllers/keystoneapi_controller.go
stuggi: We'll create the certs for public/internal in a single place and pass in a named secret to the service operator to be used. if that secret is passed (name != "") we need to set https for the internal endpoint. Thats why here is a small follow up required for TLS, in addition to pass in the certs to the deployments

controllers/keystoneapi_controller.go
```diff
@@ -452,59 +451,89 @@ func (r *KeystoneAPIReconciler) reconcileInit(
...
+			},
+		)
+
+		// Create the service
+		svc, err := service.NewService(
+			service.GenericService(&service.GenericServiceDetails{
+				Name:      endpointName,
+				Namespace: instance.Namespace,
+				Labels:    exportLabels,
+				Selector:  serviceLabels,
```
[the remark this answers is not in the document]
stuggi: this is the Selector for routing the traffic to the end pod. it is not related to the route

controllers/keystoneapi_controller.go
stuggi: I initially moved keystone.ServiceName to be defined in the api module. we can have it there and use it in both places?

api/v1beta1/keystoneapi_types.go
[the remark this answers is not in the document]
gibizer: OK then lets describe in the doc of this field that one override per endpoint type is expected.

controllers/keystoneapi_controller.go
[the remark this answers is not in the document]
gibizer: ack. I got it now.

controllers/keystoneapi_controller.go
gibizer: let's discuss this further on https://github.com/openstack-k8s-operators/openstack-operator/pull/416#discussion_r1294803990

controllers/keystoneapi_controller.go
gibizer: Could we just hide the protocol logic via the serviceOverride? So when the openstack-operator configures TLS it also adds a protocol override to the service.

api/v1beta1/keystoneapi_types.go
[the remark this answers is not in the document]
stuggi: changed to be map[string]...

stuggi:
closing in favor of #307

PR #293, opened 2023-08-01
envtest: Use latest bug fix version of k8s

kajinamit:
This is follow-up of ee391d33f9b47ec6cf0e731d80f1f9cb6a0dee96 and updates the version string so that the latest bug fix version of 1.26 release is used instead of using the specific bug fix version.


stuggi (approved):
/lgtm

PR #294, opened 2023-08-02
Skip redundant update when removing a finalizer

kajinamit:
We don't have to update a CR when the CR does not contain the finalizer being removed.


abays (approved):
/lgtm

PR #298, opened 2023-08-07
Create separate secrets for token keys and credential keys

kajinamit:
... so that we can support more than two keys without modifying the logic.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -684,7 +684,18 @@ func (r *KeystoneAPIReconciler) reconcileNormal(ctx context.Context, instance *k
 	// Create secret holding fernet keys (for token and credential)
 	//
 	// TODO key rotation
-	err = r.ensureFernetKeys(ctx, instance, helper, &configMapVars)
+	err = r.ensureFernetKeys(ctx, instance, instance.Name+"-fernet", helper, &configMapVars)
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.ServiceConfigReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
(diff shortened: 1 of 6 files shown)


controllers/keystoneapi_controller.go
```diff
@@ -1022,16 +1034,13 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
 	secret, hash, err := oko_secret.GetSecret(ctx, helper, secretName, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return err
 	} else if k8s_errors.IsNotFound(err) {
 		fernetKeys := map[string]string{
-			"FernetKeys0":     keystone.GenerateFernetKey(),
-			"FernetKeys1":     keystone.GenerateFernetKey(),
-			"CredentialKeys0": keystone.GenerateFernetKey(),
-			"CredentialKeys1": keystone.GenerateFernetKey(),
+			"0": keystone.GenerateFernetKey(),
```
fmount, line 1042: is there a particular reason to prefer "0" or "1" over a `<KeyName>\d` format?

kajinamit: "0" or "1" matches the file name expected by keystone (the key files should be placed at `/etc/keystone/fernet-keys/<index>`). This change is required to avoid maintaining mapping of keyname and the exposed path.

fmount: ok I see. I asked because given this is part of a secret, which is dumped by must-gathering where sensitive data are masked [1], this change will require an update in the `mask` script to make sure we don't miss this kind of info that shouldn't be dumped or should be masked.

[1] https://github.com/openstack-k8s-operators/openstack-must-gather/blob/main/pyscripts/mask.py#L17

PR #301, opened 2023-08-23
Do not multiply time.Second by time.Duration

kajinamit:
time.Second already represents duration. We don't have to use time.Duration to generate n seconds but can use a normal integer.


abays (approved):
/lgtm

PR #304, opened 2023-08-29
Unpin openstack-k8s-operators

kajinamit:
We pinned these to the released versions for the dev preview release but we should resume tracking the latest code.


abays (approved):
/lgtm

PR #307, opened 2023-09-07
Update keystoneapi to use service override

stuggi:
Removes creation of routes for keystoneapi as those get done in the openstack-operator. Via service overrides the parts of the service can be customized. The service operator adds annotation to the service which needs to be exposed as a route.

Jira: [OSP-26690](https://issues.redhat.com//browse/OSP-26690)

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/332


abays (approved):
/lgtm

PR #311, opened 2023-09-22
Move keystone test helpers from lib-common to /api

gibizer:


api/go.mod
```diff
@@ -3,23 +3,24 @@ module github.com/openstack-k8s-operators/keystone-operator/api
 go 1.19
 
 require (
+	github.com/go-logr/logr v1.2.4
+	github.com/google/uuid v1.3.1
+	github.com/onsi/gomega v1.27.10
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.1.1-0.20230919113507-d74c2f31d216
 	github.com/openstack-k8s-operators/lib-common/modules/openstack v0.1.1-0.20230919113507-d74c2f31d216
+	github.com/openstack-k8s-operators/lib-common/modules/test v0.1.1-0.20230919113507-d74c2f31d216
```
gibizer, line 11: We need to consider if this is an acceptable dependency growth in the /api module or not. If it is considered too much then we can put the code into /test/helpers where "helpers" will be a new go.mod to import.

If every project that depends on keystone-operator/api today also expected to have some EnvTest coverage then this dependency increase is not a big deal as those project needs these deps anyhow to implement the tests. So the real question is: are there any projects that depends on (or will depend on) keystone/api but does not want / need EnvTest coverage? Because for them this might be an unnecessary set of dependencies.

dprince: As the one who has probably been the most outspoken about the dependency growth here I think this is acceptable. One question on modules/test though below...

api/test/helpers/keystone.go
```diff
@@ -0,0 +1,252 @@
...
+
+	th.Logger.Info("KeystoneAPI created", "KeystoneAPI", name)
+	return name
+}
+
+// CreateKeystoneAPIWithFixture creates a KeystoneAPI CR and configures
+// its endpoints to point to the KeystoneAPIFixture that simulate the
+// keystone-api behavior.
+func (th *TestHelper) CreateKeystoneAPIWithFixture(
+	namespace string, fixture *apis.KeystoneAPIFixture,
```
dprince, line 93: Should the fixture here also move into the API package?

gibizer: Good point. I did not caught it as KeystoneAPIFixture does not have any keystone-operator dependency. But logically that fixture should move too. I will respin

api/go.mod
stuggi: For now I agree its ok. Depending on what is required in the test helpers in the future it could grow and we'd have to make a decision later when its to much to not be in /api module.

api/go.mod
```diff
@@ -71,3 +74,5 @@ require (
 // mschuppert: map to latest commit from release-4.13 tag
 // must consistent within modules and service operators
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
+
+replace github.com/openstack-k8s-operators/lib-common/modules/test => github.com/gibizer/lib-common/modules/test v0.0.0-20230922161820-e6fe1f3e921b
```
stuggi, line 78: is this required in this PR?

go.mod
```diff
@@ -5,15 +5,15 @@ go 1.19
...
-	github.com/openstack-k8s-operators/lib-common/modules/openstack v0.1.1-0.20230919113507-d74c2f31d216
-	github.com/openstack-k8s-operators/lib-common/modules/test v0.1.1-0.20230919113507-d74c2f31d216
-	github.com/openstack-k8s-operators/mariadb-operator/api v0.1.1-0.20230913081601-9e4fc8aadad5
+	github.com/openstack-k8s-operators/infra-operator/apis v0.1.1-0.20230920125017-2c76cd203b44
+	github.com/openstack-k8s-operators/keystone-operator/api v0.1.1-0.20230920085319-92ae0260bbf3
+	github.com/openstack-k8s-operators/lib-common/modules/common v0.1.1-0.20230925103424-7dcb605ea0bb
+	github.com/openstack-k8s-operators/lib-common/modules/database v0.1.1-0.20230925103424-7dcb605ea0bb
+	github.com/openstack-k8s-operators/lib-common/modules/openstack v0.1.1-0.20230925103424-7dcb605ea0bb
+	github.com/openstack-k8s-operators/lib-common/modules/test v0.1.2-0.20230925103424-7dcb605ea0bb
+	github.com/openstack-k8s-operators/mariadb-operator/api v0.1.1-0.20230918111825-8999b3b2dc3c
```
gibizer, line 16: The lib-common bump pulled in openstack-k8s-operator bumps as well

api/go.mod
gibizer: removed as #340 landed

gibizer:
/unhold
The dependency landed

stuggi (approved):
/lgtm

PR #317, opened 2023-09-29
[envtest]Use infra helpers from infra-operator

gibizer:
This is necessary to remove a dependency cycle from lib-common


fmount (approved):
/lgtm

PR #323, opened 2023-10-02
Use database helpers from mariadb-operator/api

gibizer:
The lib-common/modules/database is moved to mariadb-operator/api to remove a circular dependency.


stuggi (approved):
/lgtm

PR #325, opened 2023-10-05
Support for RabbitMQ notifications

mauricioharley:
Depends-on: https://github.com/openstack-k8s-operators/install_yamls/pull/655

### DIFF
--- a/config/samples/keystone_v1beta1_keystoneapi.yaml
+++ b/config/samples/keystone_v1beta1_keystoneapi.yaml
@@ -3,7 +3,6 @@ kind: KeystoneAPI
 metadata:
   name: keystone
 spec:
-  containerImage: quay.io/podified-antelope-centos9/openstack-keystone:current-podified
   adminProject: admin
   adminUser: admin
   customServiceConfig: |
--- a/config/samples/keystone_v1beta1_keystoneapi.yaml
+++ b/config/samples/keystone_v1beta1_keystoneapi.yaml
@@ -3,6 +3,7 @@ kind: KeystoneAPI
 metadata:
   name: keystone
 spec:
+  containerImage: quay.io/podified-antelope-centos9/openstack-keystone:current-podified
   adminProject: admin
   adminUser: admin
   customServiceConfig: |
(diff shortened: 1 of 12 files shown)


dprince:
It looks like the kuttl tests might need an update here?

config/samples/keystone_v1beta1_keystoneapi.yaml
```diff
@@ -3,6 +3,7 @@ kind: KeystoneAPI
 metadata:
   name: keystone
 spec:
+  containerImage: quay.io/podified-antelope-centos9/openstack-keystone:current-podified
```
abays, line 6: We shouldn't add explicit `containerImage`s in the samples, as we handle these defaults elsewhere

tests/functional/keystoneapi_controller_test.go
```diff
@@ -149,9 +150,15 @@ var _ = Describe("Keystone controller", func() {
...
 				ConditionGetterFunc(KeystoneConditionGetter),
-				condition.MemcachedReadyCondition,
+				keystonev1.KeystoneRabbitMQTransportURLReadyCondition,
 				corev1.ConditionFalse,
 			)
+			th.ExpectCondition(
+				keystoneApiName,
+				ConditionGetterFunc(KeystoneConditionGetter),
+				condition.MemcachedReadyCondition,
+				corev1.ConditionUnknown,
```
abays, line 160: It looks like `condition.MemcachedReadyCondition` used to be `corev1.ConditionFalse`.  Is there a particular reason it has been changed to `corev1.ConditionUnknown`?

xek: This is due to the order of execution. Because the setting of the condition.MemcachedReadyCondition is after KeystoneRabbitMQTransportURLReadyCondition, it never gets set, because it will exit earlier with an error

See the line where MemcachedReadyCondition is set and earlier lines here: https://github.com/openstack-k8s-operators/keystone-operator/pull/325/files/4ed509ca5748368853561e6f547235ca2bdc2701#diff-89511165055af64058394ac61e20f832806f3f325b9cb3771f723d7e6cb94c8cR709

Mauricio will add a test to execute this portion.

abays: Ack, thank you for the explanation

xek (approved):
LGTM

abays (approved):
/lgtm

PR #330, opened 2023-10-12
Switch to reusable build workflow

dprince:


.github/workflows/build-keystone-operator.yaml
```diff
@@ -11,177 +11,10 @@ env:
...
-    - name: Push keystone-operator-index To ${{ env.imageregistry }}
-      uses: redhat-actions/push-to-registry@v2
-      with:
-        image: keystone-operator-index
-        tags: ${{ env.latesttag }} ${{ github.sha }}
-        registry:  ${{ env.imageregistry }}/${{ env.imagenamespace }}
-        username: ${{ secrets.QUAY_USERNAME }}
-        password: ${{ secrets.QUAY_PASSWORD }}
+  call-build-workflow:
+    uses: dprince/openstack-k8s-operators-ci/.github/workflows/reusable-build-operator.yaml@reusable-build-workflows
```
dprince, line 15: FIXME: switch to openstack-k8s-operators once we land that PR

.github/workflows/build-keystone-operator.yaml
```diff
@@ -11,177 +11,10 @@ env:
...
-        registry:  ${{ env.imageregistry }}/${{ env.imagenamespace }}
-        username: ${{ secrets.QUAY_USERNAME }}
-        password: ${{ secrets.QUAY_PASSWORD }}
+  call-build-workflow:
+    uses: openstack-k8s-operators/openstack-k8s-operators-ci/.github/workflows/reusable-build-operator.yaml@main
+    with:
+      operator_name: keystone
+      go_version: 1.19.x
+      operator_sdk_version: 1.31.0
+    secrets: inherit
```
dprince, line 20: Hit an issue here where secrets aren't inherited across an org. So if the workflow above is in 'openstack-k8s-operators' it doesn't allow my personal 'dprince' secrets to be passed into that workflow unless I modify the uses: line above to be from my own repo (this is how I had it for testing and only just noticed it) @viroel @stuggi @gibizer Looking into options to support this for "forked" workflows. Landing this now wouldn't break our upstream builds, but it would definately impact those of us use have secrets set in our forked repositories

dprince: Okay, Figured it out. Just need to be explicit and not use 'inherit'. So this works`    secrets:
      IMAGENAMESPACE: ${{ secrets.IMAGENAMESPACE }}
      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
      REDHATIO_USERNAME: ${{ secrets.REDHATIO_USERNAME }}
      REDHATIO_PASSWORD: ${{ secrets.REDHATIO_PASSWORD }}
`

stuggi (approved):
/lgtm

abays (approved):
/lgtm

PR #350, opened 2023-12-13
add operator-sdk to localbin

yazug:
specify version for operator-sdk for make bundle
download operator-sdk pinned version when generating bundle update make bundle to use local operator-sdk


abays (approved):
/lgtm

PR #358, opened 2024-01-26
Update Dockerfile images

afaranha:
Update Dockerfile to use fips compatible images and enable CGO_ENABLED.


afaranha:
The changes this PR does was already merged in another PR

PR #359, opened 2024-01-26
[go.mod]Bump from v0.3.0 to pseudoversion

gibizer:
Our intention is to track service operator and lib-common dependencies via pseudoversions. However renovate does not automatically bump from a tagged version (e.g. v0.3.0) to a newer but not tagged pseudoversion. So this patch does the manual bump. After this renovate will bump the newer pseduoversions.


stuggi (approved):
/lgtm

PR #361, opened 2024-02-02
Drop Debug fields

gibizer:
These fields were redundant. Crashing pods can be debugged via oc debug.

Implements: [OSPRH-4290](https://issues.redhat.com//browse/OSPRH-4290)


stuggi (approved):
/lgtm

PR #366, opened 2024-02-08
bump mariadb dependency and fix envtest

stuggi:


gibizer (commented):
I think we need more. We need to have RBAC rules to manage MariaDBAccount CRs, and we probably need to declare that KeystoneAPI Owns the MariaDBAccount CR created not just the MariaDBDatabas CR. See https://github.com/openstack-k8s-operators/nova-operator/pull/660

stuggi:
> I think we need more. We need to have RBAC rules to manage MariaDBAccount CRs, and we probably need to declare that KeystoneAPI Owns the MariaDBAccount CR created not just the MariaDBDatabas CR. See [openstack-k8s-operators/nova-operator#660](https://github.com/openstack-k8s-operators/nova-operator/pull/660)

right, thanks for the pointer. I have added the rbac and a test for the finalizer. is the nova one missing `mariadbaccounts/finalizers` ?

controllers/keystoneapi_controller.go
gibizer, line 1: I would also add an Owns(MariaDBAccount) statement to https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L324

tests/functional/keystoneapi_controller_test.go
```diff
@@ -342,6 +342,11 @@ var _ = Describe("Keystone controller", func() {
 					},
 				),
 			)
+
+			apiAcc := mariadb.GetMariaDBAccount(keystoneApiName)
+			Expect(apiAcc.Finalizers).To(ContainElement("KeystoneAPI"))
```
gibizer, line 347: I'm wondering if we have / need a test case that test the removal of such finalizer.

stuggi: yes, we could. But we might want to do this is a central place in the mariadb-operator, for testing the DeleteFinalizer() function instead of in each operator?

gibizer: I had bugs during creating the bump in placement regarding what name to be used when looking up the db to call db.DeleteFinalizer(). So I think a service operator specific test is worth the effort.

stuggi: ack, I have added a test "removes the finalizers when deleted" which checks before deleting the keystoneapi that the finalizers exist and afterwards that they are gone.

gibizer (approved):
Thanks @stuggi !

gibizer:
The kuttl error is relevant
```
     case.go:366: resource KeystoneAPI:keystone-kuttl-tests/keystone: .status.databaseHostname: value mismatch, expected: openstack != actual: openstack.keystone-kuttl-tests.svc 
```
the TLS change in mariadb causing a different hostname.

gibizer (approved):

gibizer (approved):

PR #370, opened 2024-02-12
Fix reconcileDelete during namespace delete

gibizer:
`oc delete namespace openstack` hangs forever because KeystoneAPI is reconciled in the state when the namespace already being terminated but the KeystoneAPI CR has no delete timestamp set yet. In this state KeystoneAPI's reconciler tries to manipulate things in the namespace (service account, TransportURL, etc) and that is rejected by k8s. This cause KeystoneAPI to become Ready=False. This then prevent KeystoneEndpoint and KeystoneService deletion to proceed as they are waiting for KeystoneAPI to become Ready=true.

This patch implements one possible solution. It assumes if the KeystoneAPI is being deleted then there is no point removing the endpoints and users from the keystone DB as the DB will be deleted too. Therefore the KeystoneEndpoint and KeystoneService simply removes the finalizers if they detect that KeystoneAPI deleted timestamp is set.

This partially resolves the namespace deletion hangs. (There is another issue with DNSMasq CR that still blocks the namespace deletion)

Fixes: [OSP-27793](https://issues.redhat.com//browse/OSP-27793)

### DIFF
--- a/controllers/keystoneservice_controller.go
+++ b/controllers/keystoneservice_controller.go
@@ -170,6 +170,15 @@ func (r *KeystoneServiceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	// If both the service and the KeystoneAPI is deleted then we can skip
+	// the cleanup of the service in the DB as the DB is going away as well.
+	// Moreover if KeystoneAPI is being deleted then we cannot talk to the
+	// keystone REST API any more. This happens for example during namespace
+	// deletion.
+	if !instance.DeletionTimestamp.IsZero() && !keystoneAPI.DeletionTimestamp.IsZero() {
+		return r.reconcileDeleteFinalizersOnly(ctx, instance, helper, keystoneAPI)
+	}
+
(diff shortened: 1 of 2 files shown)


controllers/keystoneservice_controller.go
```diff
@@ -170,6 +170,15 @@ func (r *KeystoneServiceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	// If both the endpoint and the KeystoneAPI is deleted then we can skip
+	// the cleanup of the endpoint in the DB as the DB is going away as well.
+	// Moreover if KeystoneAPI is being deleted then we cannot talk to the
+	// keystone REST API any more. This happens for example during namespace
+	// deletion.
```
abays, line 177: Maybe the comment here should speak of `service` instead of `endpoint`?

gibizer: yepp. Copy-paste :(

gibizer:
Together with https://github.com/openstack-k8s-operators/infra-operator/pull/178 this fixes the hanging namespace deletion

abays (approved):
/lgtm

PR #371, opened 2024-02-12
migrate from databaseUsername to databaseAccount and fully use MariaDBAccount 

zzzeek:
This moves keystone  to fully use MariaDBAccount based on the dev work being done for mariadb-operator:

https://github.com/openstack-k8s-operators/mariadb-operator/pull/184/files

Lead Jira: [OSPRH-4095](https://issues.redhat.com/browse/OSPRH-4095)

1. [x] controller calls EnsureMariaDBAccount up front to make sure MariaDBAccount is present
2. [x] error code from EnsureMariaDBAccount is handled, Conditions are amended when error is returned
3. [x] controller calls NewDatabaseForAccount instead of NewDatabase
4. [x]  GetAccountAndSecret is used to retrieve account /secret to populate template
5. [x]  GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. [x]  CreateOrPatchAll() used to patch the Database, replacing CreateOrPatchDB / CreateOrPatchDBByName
7. [x]  controller calls DeleteUnusedMariaDBAccountFinalizers when launched pods are definitely running on a new MariaDBAccount, returns error code if present
8. [x]  PasswordSelectors that refer to database are removed
9. [x]  all databaseUser replaced with databaseAccount inside of all XYZ_types.go
10. [x] all databaseUser replaced with databaseAccount inside of all `kuttl/*.yaml`
11. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all XYZ_types.go
12. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all `kuttl/*.yaml`
13. [x] MariaDBAccountSuiteTests are used in controller ginkgo tests if it has them
14. [x] [184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) is merged and replaces from go.mod are removed


controllers/keystoneapi_controller.go
```diff
@@ -880,6 +880,21 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
...
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err = mariadbv1.EnsureMariaDBAccount(
+		ctx, helper, instance.Spec.DatabaseAccount,
+		instance.Namespace, false,
+	)
+
+	if err != nil {
+		return ctrl.Result{}, err
```
stuggi, line 895: it might be good to have a condition which reflects the status of the MariaDBAccount. if not it would allow to see where it is stuck in reconcile.

controllers/keystoneapi_controller.go
```diff
@@ -880,6 +880,21 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
...
 	//
 
+	// ensure MariaDBAccount exists.  This account record may be created by
+	// openstack-operator or the cloud operator up front without a specific
+	// MariaDBDatabase configured yet.   Otherwise, a MariaDBAccount CR is
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err = mariadbv1.EnsureMariaDBAccount(
+		ctx, helper, instance.Spec.DatabaseAccount,
```
stuggi, line 890: iiuc if the DB password changes there will be a new `DatabaseAccount` with a new name. Is there some situation where the current DatabaseAccount changes and the service operator has to watch this named resource similar to https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L248-L258 ?

zzzeek: the MariaDBAccount object itself, as well as its associated Secret is never meant to change (i.e. the secret is ideally immutable), the mariadbaccount_controller does not respond to changes in the MariaDBAccount/secret itself.  there's only CREATE and DROP for the creation and dropping of the CR.

controllers/keystoneapi_controller.go
```diff
@@ -1197,8 +1215,8 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
 		"memcachedServers": strings.Join(mc.Status.ServerList, ","),
 		"TransportURL":     string(transportURLSecret.Data["transport_url"]),
 		"DatabaseConnection": fmt.Sprintf("mysql+pymysql://%s:%s@%s/%s",
-			instance.Spec.DatabaseUser,
-			string(ospSecret.Data[instance.Spec.PasswordSelectors.Database]),
+			databaseAccount.Spec.UserName,
+			string(dbSecret.Data["DatabasePassword"]),
```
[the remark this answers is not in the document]
stuggi: yes I think it makes sense to have a shared const for this

controllers/keystoneapi_controller.go
zzzeek: OK, I would add this to all the operators I'm doing...

zzzeek: hmm glance-operator does the conditions differently, it already applies an error for this in its conditions, just not a specific one.    guess this will have to be case by case

zzzeek:
@stuggi this one is ready to merge

api/v1beta1/keystoneapi_types.go
```diff
@@ -121,7 +120,7 @@ type KeystoneAPISpecCore struct {
 	TrustFlushSuspend bool `json:"trustFlushSuspend"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={database: KeystoneDatabasePassword, admin: AdminPassword}
+	// +kubebuilder:default={admin: AdminPassword}
 	// PasswordSelectors - Selectors to identify the DB and AdminUser password from the Secret
```
stuggi, line 124: nit: update comment to remove reference to DB password

controllers/keystoneapi_controller.go
```diff
@@ -1048,6 +1048,14 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
 	}
 	// create Deployment - end
 
+	if instance.Status.ReadyCount > 0 {
```
stuggi, line 1051: should this be  `instance.Status.ReadyCount == instance.Spec.Replicas`? like all replicas are up using the new account, then we can remove the finalizer.

controllers/keystoneapi_controller.go
```diff
@@ -1381,24 +1387,46 @@ func (r *KeystoneAPIReconciler) ensureDB(
...
+
+	// ensure MariaDBAccount exists.  This account record may be created by
+	// openstack-operator or the cloud operator up front without a specific
+	// MariaDBDatabase configured yet.   Otherwise, a MariaDBAccount CR is
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err := mariadbv1.EnsureMariaDBAccount(
+		ctx, h, instance.Spec.DatabaseAccount,
+		instance.Namespace, false, "keystone",
```
stuggi, line 1399: in a follow up we might want to change where `"keystone"` or `instance.Name` is used for DB and DB CR to be `keystone.DatabaseName`, like e.g. in `NewDatabaseForAccount` and remove finalizer cleanup

controllers/keystoneapi_controller.go
zzzeek: OK, sounds good

stuggi (approved):
/lgtm as discussed doing improvements in follow up

controllers/keystoneapi_controller.go
zzzeek: I haven't decided on a consistent approach for these however in Ironic I have DatabaseName and DatabaseCRName, both constants.

I've seen instance.Name used quite a lot but what's the use case for instance.Name being copied as the database name, and if so why not the CR name?        What's the argument for "the DB name / CR name is constant" vs. "it is dynamic on instance.Name" - we aren't being consistent about this across projects and I think "they are both constants" is likely best

zzzeek: obviously not for nova cell DBs, just for the usual case of ironic / cinder / keystone / neutronapi /eetc.   in neutron's case we have to use a constant since apparently they have some kind of dynamic scheme for instance.name

zzzeek: the "keystone" name in this particular call is a prefix for creating MariaDB usernames.  it is not ever keyed after this as it's not a complete name.

controllers/keystoneapi_controller.go
zzzeek: see https://github.com/openstack-k8s-operators/keystone-operator/pull/390

PR #373, opened 2024-02-15
Create database before config rendering

stuggi:
dbsync can run without proper DB connection url configured when there is no reconcile event happening and the service config won't get updated before the job gets created.
This moves the DB create handling before calling `generateServiceConfig...`


gibizer (approved):
Looks good

PR #391, opened 2024-03-07
Fix k8s.io/kube-openapi dep

gibizer:
The k8s.io/kube-openapi v0.0.0-20240126223410-2919ad4fcfec indirect
dependency needs golang 1.21 so we should not use it. Recently that
package reverted to use golang 1.20. So instead of downgrading this PR
is upgrading to the latest.


gibizer:
Note that we fixed this before but the wrong dep was reintroduced recently

abays (approved):
/lgtm

PR #393, opened 2024-03-13
update mariadb to include finalizer secret logic

zzzeek:
this PR updates mariadb-operator to include [1], which will add / remove finalizers on the Secret objects that are associated with MariaDBAccount objects.  This ensures that GetDatabaseByNameAndAccount will successfully locate a MariaDBDatabase/MariaDBAccount/Secret trio, rather than returning not found, allowing ``Database.DeleteFinalizer()`` to be called when a resource is in reconcile delete.

[1] https://github.com/openstack-k8s-operators/mariadb-operator/pull/210

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.3.1-0.20240326081751-56015b1ae3f6
 	github.com/openstack-k8s-operators/lib-common/modules/openstack v0.3.1-0.20240216173409-86913e6d5885
 	github.com/openstack-k8s-operators/lib-common/modules/test v0.3.1-0.20240216173409-86913e6d5885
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

zzzeek:
keystone is still behind, this needs to merge

abays:
@zzzeek Looks like KUTTL tests need to be updated:

```
case.go:366: resource KeystoneAPI:keystone-kuttl-tests/keystone: .metadata.finalizers: slice length mismatch: 2 != 3
```

zzzeek:
that failure is a little bit odd.   i will add that finalizer and see if it helps.   but...might need to understand what's going on there

zzzeek:
i think that finalizer is not supposed to be there.  let me do a retest and if that fails ill try to run it all locally

zzzeek:
Ok I guess the previous fail was a timeout situation

PR #395, opened 2024-03-22
Enforcing scope with SRBAC breaks heat

rabi:
Heat won't work when scope is enforced and it's being used by many of our NFV customers. Rather than making a single property to enable SRBAC, let's split them so that `enforce_new_defaults` can be set to true by default and customers can toggle `enforce_scope` if they're not using heat.

jira: https://issues.redhat.com/browse/OSPRH-5753

### DIFF
--- a/api/bases/keystone.openstack.org_keystoneapis.yaml
+++ b/api/bases/keystone.openstack.org_keystoneapis.yaml
@@ -84,11 +84,6 @@ spec:
                   files. Those get added to the service config dir in /etc/<service>
                   . TODO: -> implement'
                 type: object
-              enableSecureRBAC:
-                default: true
-                description: EnableSecureRBAC - Enable Consistent and Secure RBAC
-                  policies
-                type: boolean
               memcachedInstance:
                 default: memcached
                 description: Memcached instance name.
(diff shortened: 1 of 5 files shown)


rabi:
I've mentioned it in the jira and doing it here too.. The other option is changing keystone default policies[1] to allow domain scope.

[1] https://github.com/openstack/keystone/blob/master/keystone/common/policies/role.py#L98

vyzigold:
I'll note it here. I just tried to create a stack and it was failing with the following until I disabled SRBAC:
```
Resource CREATE failed: Forbidden: resources.scaledown_policy: You are not authorized to perform the requested action: identity:list_roles.
```

My hopes are this PR will fix the issue.

rabi:
> My hopes are this PR will fix the issue.

Toggling enforce_scope will fix the issue as explained in the commit message. You can test with the `customServiceConfig` as below for keystone.

  ```
keystone:
    apiOverride:
      route: {}
    enabled: true
    template:
      adminProject: admin
      adminUser: admin
      containerImage: ""
      customServiceConfig: |
        [oslo_policy]
        enforce_scope=false
```

abays:
This looks good to me.  @dmendiza What do you think?

abays (approved):
We at least need this now to unblock other things.  If there are objections, we can discuss in a follow-up.

/lgtm

PR #402, opened 2024-04-12
Add funcs for returning a list of endpoints and a list of endpointurls

jlarriba:


api/v1beta1/keystoneendpoint.go
```diff
@@ -190,3 +190,42 @@ func DeleteKeystoneEndpointWithName(
...
+	ke := &KeystoneEndpointList{}
+	err := h.GetClient().List(ctx, ke)
+	if err != nil {
+		return nil, err
+	}
+
+	return ke, nil
+}
+
+// GetKeystoneEndpointUrls returns all URLs currently registered
```
stuggi, line 210: can you please add a description of what visibility is  used for, e.g. like the `--interface` from the openstackclient, could also rename interface to call it like the openstackclient.

```
  --interface <interface>
                        Filter by interface type (admin, public or internal)
```

 also do we want to do this e.g. a string ptr and if it is nil, we return all of them and not filter on visibility?

jlarriba:
/retest

`Get "https://api.hosted-mgmt.ci.devcluster.openshift.com:6443/apis/hive.openshift.io/v1": dial tcp: lookup api.hosted-mgmt.ci.devcluster.openshift.com on 172.30.0.10:53: no such host`

api/v1beta1/keystoneendpoint.go
```diff
@@ -190,3 +190,52 @@ func DeleteKeystoneEndpointWithName(
...
+
+// GetKeystoneEndpointList func
+func GetKeystoneEndpointList(
+	ctx context.Context,
+	h *helper.Helper,
+	namespace string,
+) (*KeystoneEndpointList, error) {
+
+	ke := &KeystoneEndpointList{}
+	err := h.GetClient().List(ctx, ke)
```
stuggi, line 202: its not filtered for the namespace, or ? should probably use some list option:

```
        listOpts := []client.ListOption{
                client.InNamespace(namespace),
        }

        if err := h.GetClient().List(ctx, ke, listOpts...); err != nil {
                err = fmt.Errorf("Error listing endpoints for namespace %s: %w", namespace, err)
		return nil, err
        }
```

api/v1beta1/keystoneendpoint.go
```diff
@@ -190,3 +190,52 @@ func DeleteKeystoneEndpointWithName(
...
+// If it is nil, all URLs are returned
+func GetKeystoneEndpointUrls(
+	ctx context.Context,
+	h *helper.Helper,
+	namespace string,
+	visibility *string,
+) ([]string, error) {
+
+	ke := &KeystoneEndpointList{}
+	err := h.GetClient().List(ctx, ke)
```
stuggi, line 220: same question on namespace filtering, could just call GetKeystoneEndpointList()

stuggi (changes requested):

api/v1beta1/keystoneendpoint.go
jlarriba: Absolutely right, sir.

api/v1beta1/keystoneendpoint.go
jlarriba: Yes indeed, we should be calling GetKeystoneEndpointList

stuggi (approved):
/lgtm

PR #403, opened 2024-04-15
Use system scoped admin service client per default

stuggi:
To customize roles for a user, like assigining "SwiftSystemReader" role to the swift user in the "service" project within the telemetry operator a system scoped client is required. Since the client is used for deployment its probably ok to always use a system scoped client.

Also adds GetScopedAdminServiceClient() if an admin cliend with a different scope is required.

### DIFF
--- a/api/v1beta1/keystoneapi.go
+++ b/api/v1beta1/keystoneapi.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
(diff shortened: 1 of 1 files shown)


api/v1beta1/keystoneapi.go
```diff
@@ -69,11 +70,33 @@ func GetKeystoneAPI(
...
+
+// GetScopedAdminServiceClient - get a scoped admin serviceClient for the keystoneAPI instance
+func GetScopedAdminServiceClient(
+	ctx context.Context,
+	h *helper.Helper,
+	keystoneAPI *KeystoneAPI,
+	scope *gophercloud.AuthScope,
 ) (*openstack.OpenStack, ctrl.Result, error) {
 	// get public endpoint as authurl from keystone instance
 	authURL, err := keystoneAPI.GetEndpoint(endpoint.EndpointInternal)
```
jlarriba, line 102: The following parameters:

```
Username:   keystoneAPI.Spec.AdminUser,
TenantName: keystoneAPI.Spec.AdminProject,
```

are not needed here to choose the admin user in the admin  project?

stuggi: not sure I understand, the AdminUser and Project is used L138/L140. you refer to that one?

jlarriba: oh yes, it was outside of the changes so i didnt realized it was already being done.

PR #408, opened 2024-04-24
Bump keystone probe timeouts

SeanMooney:
When deploying locally in crc keystone can crashloop
indefintly because the readiness and liveness probe
were set to strictly.

The nova-operator relaxed the probes becuase of tempest failures
in openstack-k8s-operators/nova-operator#381

This change updates the keystone values to match.


stuggi (approved):
/lgtm

PR #414, opened 2024-05-07
Check for valid service override endpoint type on create and update

stuggi:
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/505


xek (approved):
LGTM

PR #432, opened 2024-06-27
Create cloudrc contents in openstack-config-secret

Akrog:
We are currently only creating the `clouds.yaml` and `secure.yaml` for the openstack client and not the classic `cloudrc` file that was usually sourced for individual clients.

This makes sense if there were parity between the openstck client and all the individual clients, but there are still missing features. For example:

- The `openstack volume service list` cannot show the cluster.
- The `openstack volume service list` cannot show the backend state.
- There is no support for any of the default volume type commands: set, show, list, unset.

These are some of the ones we know about, but there could be more.

Users will need to use the old `cinder` client, so we should make it convenient for them by having the `cloudrc` available in the `openstackclient` pod.

This patch adds the `cloudrc` key in the `openstack-config-secret` secret to make it possible for the `openstack-operator` to mount it in the `openstackclient` pod.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -1280,8 +1280,10 @@ func (r *KeystoneAPIReconciler) reconcileCloudConfig(
 	if err != nil {
 		return err
 	}
+	cloudrc := keystone.GenerateCloudrc(&openStackConfigSecret, &openStackConfig)
 	secretString := map[string]string{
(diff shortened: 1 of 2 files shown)


ASBishop (approved):
/lgtm

ASBishop:
@frenzyfriday can you please approve this? It's needed for https://github.com/openstack-k8s-operators/openstack-operator/pull/890

PR #455, opened 2024-07-30
Update reconcileNormal to use VerifySecret

mauricioharley:


controllers/keystoneapi_controller.go
```diff
@@ -712,15 +712,15 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
-	ospSecret, hash, err := oko_secret.GetSecret(ctx, helper, instance.Spec.Secret, instance.Namespace)
+	hash, result, err := oko_secret.VerifySecret(ctx, types.NamespacedName{Name: instance.Spec.Secret, Namespace: instance.Namespace}, []string{"AdminPassword"}, helper.GetClient(), time.Second*10)
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
```
abays, line 717: I think since the `IsNotFound` error is handled in `lib-common` [1], this should be checking `result != ctrl.Result{}` instead.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/7fd3da600c9d59fd81e86f32893aaadece9cc856/modules/common/secret/secret.go#L422-L426

abays: What I mean by that is have an `else` for the `if err != nil` that checks `result != ctrl.Result{}`.

controllers/keystoneapi_controller.go
```diff
@@ -908,6 +914,11 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
 			err.Error()))
 		return ctrlResult, err
 	} else if (ctrlResult != ctrl.Result{}) {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.TLSInputReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			fmt.Sprintf(condition.TLSInputReadyWaitingMessage, "one or more cert secrets")))
```
abays, line 921: @stuggi I wanted to get your input here.  The call to `ValidateCertSecrets` on L907 no longer returns an error with the name of the missing TLS `Secret` because we (I) updated `lib-common` to not return an error when we're also returning a `RequeueAfter` [1].  So here we don't know which TLS cert `Secret` is missing, hence we're using `"one or more cert secrets"` as the requeue status message.  Just wondering if you think we should refactor the chain from `ValidateCertSecrets` through `VerifySecret` in `lib-common` to be able to return a missing `Secret` name, so that we can be more explicit here.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/547

stuggi: we get the RequeueAfter only if the secrets are really missing. if there is an error with the content we get an error, right? So we could either do what you said, or we just have a generic message like ""TLSInput is missing: %s", where %s is  "%s and/or %s", `instance.Spec.TLS.API.Internal.SecretName`, `instance.Spec.TLS.API.Public.SecretName`"

PR #468, opened 2024-09-05
Update tls cert secret validation due VerifyCertSecrets() change

stuggi:
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/559

Jira: [OSPRH-9991](https://issues.redhat.com//browse/OSPRH-9991)


tests/functional/keystoneapi_controller_test.go
```diff
@@ -878,7 +878,8 @@ var _ = Describe("Keystone controller", func() {
 				condition.TLSInputReadyCondition,
 				corev1.ConditionFalse,
 				condition.RequestedReason,
-				"TLSInput is missing: one or more cert secrets",
+				fmt.Sprintf("TLSInput is missing: secrets \"%s not found in namespace %s\" not found",
```
abays, line 881: The string looks somewhat weird.  It has `not found` twice and doesn't read correctly.

stuggi: then we have to change https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/tls/tls.go#L263 . the 2nd not found is coming from the err.Error()

stuggi: could update lib common to have 

`fmt.Sprintf("%s in namespace %s", caSecret.Name, caSecret.Namespace),`

PR #477, opened 2024-10-02
[WIP] Support variable number of fernet keys

afaranha:
jira: [OSPRH-10024](https://issues.redhat.com//browse/OSPRH-10024)


controllers/keystoneapi_controller.go
```diff
@@ -1336,11 +1336,12 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
 		return err
 	} else if k8s_errors.IsNotFound(err) {
 		fernetKeys := map[string]string{
-			"FernetKeys0":     keystone.GenerateFernetKey(),
-			"FernetKeys1":     keystone.GenerateFernetKey(),
 			"CredentialKeys0": keystone.GenerateFernetKey(),
 			"CredentialKeys1": keystone.GenerateFernetKey(),
 		}
+		for i := range instance.Spec.FernetKeys {
```
xek, line 1342: We need to check not only if the secret exists, but also generate extra keys if they are missing. This operation shouldn't rotate the keys, so the staged key (number 0) should stay as is, but the primary (the last key on the list) and secondary keys should be moved to the tail of the list of keys.
If the list in the secret exceeds the set number of keys, the keys in the middle should be deleted, compacting it, with the first key and the tail of the list having the same keys in the same order.

xek:
These changes were merged in #478

Deydra71:
@afaranha Hello! I think we should close this, right?

PR #478, opened 2024-10-04
Add fernet key rotation cronjob

xek:
As part of this work, we needed to implement more than 2 keys, since rotating 2 would expire sessions on every rotation. There are new settings and the defaults are the same as in "old-gen" Tripelo.

jira: [OSPRH-9309](https://issues.redhat.com//browse/OSPRH-9309)

### DIFF
--- a/api/v1beta1/keystoneapi_types.go
+++ b/api/v1beta1/keystoneapi_types.go
@@ -120,6 +120,18 @@ type KeystoneAPISpecCore struct {
 	TrustFlushSuspend bool `json:"trustFlushSuspend"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	// FernetRotationDays - Rotate fernet token keys every X days
+	FernetRotationDays *int32 `json:"fernetRotationDays"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=5
+	// +kubebuilder:validation:Minimum=3
+	// FernetMaxActiveKeys - Maximum number of fernet token keys after rotation
+	FernetMaxActiveKeys *int32 `json:"fernetMaxActiveKeys"`
+
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:default={admin: AdminPassword}
 	// PasswordSelectors - Selectors to identify the AdminUser password from the Secret
 	PasswordSelectors PasswordSelector `json:"passwordSelectors"`
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -857,7 +857,6 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
 	//
 	// Create secret holding fernet keys (for token and credential)
 	//
-	// TODO key rotation
 	err = r.ensureFernetKeys(ctx, instance, helper, &configMapVars)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
@@ -1321,37 +1320,53 @@ func (r *KeystoneAPIReconciler) reconcileCloudConfig(
 	return oko_secret.EnsureSecrets(ctx, h, instance, secrets, nil)
 }
 
-// ensureFernetKeys - creates secret with fernet keys
+// ensureFernetKeys - creates secret with fernet keys, rotates the keys
 func (r *KeystoneAPIReconciler) ensureFernetKeys(
 	ctx context.Context,
 	instance *keystonev1.KeystoneAPI,
 	helper *helper.Helper,
 	envVars *map[string]env.Setter,
 ) error {
+	fernetAnnotation := labels.GetGroupLabel(keystone.ServiceName) + "/rotatedat"
 	labels := labels.GetLabels(instance, labels.GetGroupLabel(keystone.ServiceName), map[string]string{})
+	now := time.Now().UTC()
 
 	//
 	// check if secret already exist
 	//
 	secretName := keystone.ServiceName
+	var numberKeys int
+	if instance.Spec.FernetMaxActiveKeys == nil {
+		numberKeys = keystone.DefaultFernetMaxActiveKeys
+	} else {
+		numberKeys = int(*instance.Spec.FernetMaxActiveKeys)
+	}
+
 	secret, hash, err := oko_secret.GetSecret(ctx, helper, secretName, instance.Namespace)
+
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return err
 	} else if k8s_errors.IsNotFound(err) {
 		fernetKeys := map[string]string{
-			"FernetKeys0":     keystone.GenerateFernetKey(),
-			"FernetKeys1":     keystone.GenerateFernetKey(),
 			"CredentialKeys0": keystone.GenerateFernetKey(),
 			"CredentialKeys1": keystone.GenerateFernetKey(),
 		}
 
+		for i := 0; i < numberKeys; i++ {
+			fernetKeys[fmt.Sprintf("FernetKeys%d", i)] = keystone.GenerateFernetKey()
+		}
+
+		annotations := map[string]string{
+			fernetAnnotation: now.Format(time.RFC3339)}
+
 		tmpl := []util.Template{
 			{
-				Name:       secretName,
-				Namespace:  instance.Namespace,
-				Type:       util.TemplateTypeNone,
-				CustomData: fernetKeys,
-				Labels:     labels,
+				Name:        secretName,
+				Namespace:   instance.Namespace,
+				Type:        util.TemplateTypeNone,
+				CustomData:  fernetKeys,
+				Labels:      labels,
+				Annotations: annotations,
 			},
 		}
 		err := oko_secret.EnsureSecrets(ctx, helper, instance, tmpl, envVars)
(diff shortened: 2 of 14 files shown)


xek:
I'm keeping it as draft, since there is a mismatch of the names of mounted keys [1]. This will likely change, since 
keystone-operator currently supports only 2 keys and there is ongoing work to support a variable number of keys #477

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_keystone-operator/478/pull-ci-openstack-k8s-operators-keystone-operator-main-keystone-operator-build-deploy-kuttl/1842180087600910336/artifacts/keystone-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-3d1f620ae3617fbd6371f55aa948b34e9daf31a461d3c557cf1a7c2db38d0385/namespaces/openstack/cronjobs/keystone-fernet-cronjob.yaml

dprince:
If we land this PR please add the associated RELATED_IMAGE into the openstack-operator here: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/default/manager_default_images.yaml

api/v1beta1/keystoneapi_types.go
```diff
@@ -119,6 +125,16 @@ type KeystoneAPISpecCore struct {
...
 
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="1 0 * * *"
+	// FernetRotationSchedule - Schedule rotate fernet token keys
+	FernetRotationSchedule string `json:"fernetRotationSchedule"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="5"
+	// FernetMaxActiveKeys - Maximum number of fernet token keys after rotation
+	FernetMaxActiveKeys string `json:"fernetMaxActiveKeys"`
```
olliewalsh, line 136: wrong type?

api/v1beta1/keystoneapi_types.go
```diff
@@ -125,6 +125,11 @@ type KeystoneAPISpecCore struct {
 	// TrustFlushSuspend - Suspend the cron job to purge trusts
 	TrustFlushSuspend bool `json:"trustFlushSuspend"`
 
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// FernetRotationDays - Rotate fernet token keys every X days
+	FernetRotationDays *int32 `json:"fernetRotationDays"`
```
olliewalsh, line 131: I would set a minimum to 1

xek: Done, I also set the minimum number of fernet keys to 3

api/v1beta1/keystoneapi_types.go
xek: I changed it to int32

controllers/keystoneapi_controller.go
```diff
@@ -1361,9 +1369,95 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
+		extraKey := fmt.Sprintf("FernetKeys%d", numberKeys)
+
+		//
+		// Fernet Key rotation
+		//
+		rotatedAt, err := time.Parse(time.RFC3339, labels[fernetLabel])
+		duration := int(*instance.Spec.FernetRotationDays)
+
+		if err != nil {
+			labels[fernetLabel] = now.Format(time.RFC3339)
```
stuggi, line 1384: can not store that in the label. 

```
  2024-10-18T11:32:47.708+0200  ERROR   Reconciler error        {"controller": "keystoneapi", "controllerGroup": "keystone.openstack.org", "controllerKind": "Keyston
eAPI", "KeystoneAPI": {"name":"keystone","namespace":"a0512b5c-43e5-4395-8719-a0ec684c8605"}, "namespace": "a0512b5c-43e5-4395-8719-a0ec684c8605", "name": "keystone"
, "reconcileID": "c21e82c2-e40f-47ff-9a2e-a2d4eb4d3e7d", "error": "Secret \"keystone\" is invalid: metadata.labels: Invalid value: \"2024-10-18T09:32:47Z\": a valid 
label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_
value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"} 
```

as discussed store it in an annotation

controllers/keystoneapi_controller.go
```diff
@@ -1361,9 +1371,96 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
+
+		//
+		// Fernet Key rotation
+		//
+		rotatedAt, err := time.Parse(time.RFC3339, secret.Annotations[fernetAnnotation])
+		duration := int(*instance.Spec.FernetRotationDays)
+
+		if err != nil {
+			secret.Annotations[fernetAnnotation] = now.Format(time.RFC3339)
+		} else if rotatedAt.AddDate(0, 0, duration).After(now) {
```
stuggi, line 1387: I think this needs to be Before() instead of After()?

api/v1beta1/keystoneapi_types.go
```diff
@@ -119,6 +119,18 @@ type KeystoneAPISpecCore struct {
 	// TrustFlushSuspend - Suspend the cron job to purge trusts
 	TrustFlushSuspend bool `json:"trustFlushSuspend"`
 
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=0
```
olliewalsh, line 124: should that be 1? I think setting it to 0 would create an infinite reconcile loop

xek: right, I set it to 1

controllers/keystoneapi_controller.go
```diff
@@ -1361,9 +1371,97 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
+
+		if !changedKeys {
+			return nil
+		}
+
+		fernetKeys := make(map[string]string, len(secret.Data))
+		for k, v := range secret.Data {
+			fernetKeys[k] = string(v[:])
+		}
+
```
stuggi, line 1448: I think we have to update the timestamp when we got here ? `secret.Annotations[fernetAnnotation] = now.Format(time.RFC3339)`

stuggi: right now we only set it on new (NotFound) or when there is an error parsing the timestamp

controllers/keystoneapi_controller.go
```diff
@@ -1361,9 +1371,98 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
-	// TODO: fernet key rotation
+		changedKeys := false
+
+		extraKey := fmt.Sprintf("FernetKeys%d", numberKeys)
+
+		//
+		// Fernet Key rotation
+		//
+		rotatedAt, err := time.Parse(time.RFC3339, secret.Annotations[fernetAnnotation])
+		duration := int(*instance.Spec.FernetRotationDays)
```
olliewalsh, line 1383: also could be nil

controllers/keystoneapi_controller.go
```diff
@@ -1321,37 +1320,48 @@ func (r *KeystoneAPIReconciler) reconcileCloudConfig(
...
 			"CredentialKeys1": keystone.GenerateFernetKey(),
 		}
 
+		for i := 0; i < numberKeys; i++ {
+			fernetKeys[fmt.Sprintf("FernetKeys%d", i)] = keystone.GenerateFernetKey()
+		}
+
+		annotations := map[string]string{
+			fernetAnnotation: now.Format(time.RFC3339)}
+
```
stuggi, line 1356: need to initialize the map outside the if, otherwise the one in L1448 can fail, like if you go afterwards to the secret and delete the annotation after initial create

controllers/keystoneapi_controller.go
```diff
@@ -1361,9 +1371,98 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
+		if !changedKeys {
+			return nil
+		}
+
+		fernetKeys := make(map[string]string, len(secret.Data))
+		for k, v := range secret.Data {
+			fernetKeys[k] = string(v[:])
+		}
+
+		secret.Annotations[fernetAnnotation] = now.Format(time.RFC3339)
```
stuggi, line 1448: secret.Annotations can be nil. e.g. when someone manually deletes the annotation from the secret.

controllers/keystoneapi_controller.go
```diff
@@ -1361,9 +1376,107 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
+
+		//
+		// Fernet Key rotation
+		//
+		if secret.Annotations == nil {
+			secret.Annotations = map[string]string{}
+		}
+		rotatedAt, err := time.Parse(time.RFC3339, secret.Annotations[fernetAnnotation])
+
+		var duration int
```
olliewalsh, line 1392: nit: would move this up so that err is checked right after setting

controllers/keystoneapi_controller.go
```diff
@@ -1361,9 +1376,107 @@ func (r *KeystoneAPIReconciler) ensureFernetKeys(
...
+			duration = keystone.DefaultFernetRotationDays
+		} else {
+			duration = int(*instance.Spec.FernetRotationDays)
+		}
+
+		if err != nil {
+			changedKeys = true
+		} else if rotatedAt.AddDate(0, 0, duration).Before(now) {
+			secret.Data[extraKey] = secret.Data["FernetKeys0"]
+			secret.Data["FernetKeys0"] = []byte(keystone.GenerateFernetKey())
```
olliewalsh, line 1403: this doesn't look right. It moves key 0 to the end, and replaces it with the new key. Should replace it with key 1... etc... and FernetKeys<numberKeys-1> should be the new key

olliewalsh (approved):
/lgtm

PR #479, opened 2024-10-09
Add OIDC Federation Settings

d34dh0r53:
This templates the OIDC federation settings needed to configure Keystone to perform federation authentication.


controllers/keystoneapi_controller.go
```diff
@@ -1203,6 +1207,21 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
 			endptConfig["SSLCertificateKeyFile"] = fmt.Sprintf("/etc/pki/tls/private/%s.key", endpt.String())
 		}
+		if instance.Spec.OIDCFederation.EnableFederation {
+			endptConfig["EnableFederation"] = true
+			endptConfig["OIDCClaimPrefix"] = instance.Spec.OIDCFederation.OIDCClaimPrefix
+			endptConfig["OIDCResponseType"] = instance.Spec.OIDCFederation.OIDCClaimPrefix
+			endptConfig["OIDCScope"] = instance.Spec.OIDCFederation.OIDCScope
+			endptConfig["OIDCProviderMetadataUrl"] = instance.Spec.OIDCFederation.OIDCProviderMetadataURL
+			endptConfig["OIDCClientID"] = instance.Spec.OIDCFederation.OIDCClientID
+			endptConfig["OIDCClientSecret"] = instance.Spec.OIDCFederation.OIDCClientSecret
```
vakwetu, line 1217: This doesn't work like this.  Look at https://github.com/openstack-k8s-operators/keystone-operator/blob/47b983c3ab77e998ebb80b6024a9fdc6a75585fa/api/v1beta1/keystoneapi.go#L137 for an example on how to extract the passwords from osp.secret

templates/keystoneapi/config/keystone.conf
```diff
@@ -11,6 +11,14 @@ enabled=true
 memcache_servers={{ .memcachedServers }}
 tls_enabled={{ .memcachedTLS }}
 
+{{if .enableFederation }}
+[federation]
+trusted_dashboard=.federationTrustedDashboard
```
vakwetu, line 16: You need {{ }} here (and below)

api/v1beta1/keystoneapi_types.go
```diff
@@ -183,6 +185,53 @@ type PasswordSelector struct {
 	// +kubebuilder:default="AdminPassword"
 	// Admin - Selector to get the keystone Admin password from the Secret
 	Admin string `json:"admin"`
+
```
vakwetu, line 188: As these will be in osp-secret, we may want to preface these with Keystone

api/v1beta1/keystoneapi_types.go
```diff
@@ -183,6 +185,53 @@ type PasswordSelector struct {
...
+	OIDCProviderMetadataURL string `json:"oidcProviderMetadataURL,omitempty"`
+
+	// OIDCClientID
+	OIDCClientID string `json:"oidcClientID,omitempty"`
+
+	// OIDCClientSecret
+	OIDCClientSecret PasswordSelector `json:"oidcClientSecret"`
+
+	// OIDCCryptoPassphrase
+	OIDCCryptoPassphrase PasswordSelector `json:"oidcCryptoPassphrase"`
```
vakwetu, line 219: Doesn't work like this - these shouldn't be defined here

controllers/keystoneapi_controller.go
```diff
@@ -1203,6 +1207,31 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
+			endptConfig["OIDCResponseType"] = instance.Spec.OIDCFederation.OIDCClaimPrefix
+			endptConfig["OIDCScope"] = instance.Spec.OIDCFederation.OIDCScope
+			endptConfig["OIDCProviderMetadataUrl"] = instance.Spec.OIDCFederation.OIDCProviderMetadataURL
+			endptConfig["OIDCClientID"] = instance.Spec.OIDCFederation.OIDCClientID
+			endptConfig["OIDCClientSecret"], _, _ = oko_secret.GetDataFromSecret(
+				ctx,
+				h,
+				instance.Spec.PasswordSelectors.KeystoneOIDCClientSecret,
+				10*time.Second,
+				"KeystoneOIDCClientSecret")
```
vakwetu, line 1222: You are going to want to handle the returns on this.  ie. what happens if the secret does not exist?

controllers/keystoneapi_controller.go
```diff
@@ -1203,6 +1207,31 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
+			endptConfig["OIDCCryptoPassphrase"], _, _ = oko_secret.GetDataFromSecret(
+				ctx,
+				h,
+				instance.Spec.PasswordSelectors.KeystoneOIDCCryptoPassphrase,
+				10*time.Second,
+				"KeystoneOIDCCryptoPassphrase")
```
vakwetu, line 1228: ditto

api/v1beta1/keystoneapi_types.go
```diff
@@ -168,6 +168,8 @@ type KeystoneAPISpecCore struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to the TLS
 	TLS tls.API `json:"tls,omitempty"`
+
```
vakwetu, line 171: As discussed on slack, add kubeBuilder options here just like nodeSelector

api/v1beta1/keystoneapi_types.go
```diff
@@ -183,6 +185,70 @@ type PasswordSelector struct {
...
+	// OIDCCryptoPassphrase - Selector to get the OIDC crypto passphrase from the Secret
+	KeystoneOIDCCryptoPassphrase string `json:"keystoneOIDCCryptoPassphrase"`
+}
+
+type KeystoneFederationSpec struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// Enablement of Federation configuration
+	EnableFederation bool `json:"enableFederation,omitempty"`
```
vakwetu, line 200: Not needed then - just check for len(instance.spec.OIDCFederation) >0

d34dh0r53: Right now I'm going to leave the toggle for testing. It also follows the pattern of Secure RBAC in that the deployer can explicitly enable or disable it.

controllers/keystoneapi_controller.go
d34dh0r53: Done

controllers/keystoneapi_controller.go
d34dh0r53: Done

api/v1beta1/keystoneapi_types.go
vakwetu: OK - just realize of course that the reason upstream tests are failing is because you are not checking for len(instance.spec.OIDCFederation)>0

api/bases/keystone.openstack.org_keystoneapis.yaml
```diff
@@ -105,6 +109,53 @@ spec:
                 description: NodeSelector to target subset of worker nodes running
                   this service
                 type: object
+              oidcFederation:
+                description: KeystoneFederationSpec to provide the configuration values
+                  for OIDC Federation
+                properties:
+                  KeystoneFederationIdentityProviderName:
```
d34dh0r53, line 116: This should be lowercase K

api/v1beta1/keystoneapi_types.go
```diff
@@ -183,6 +192,70 @@ type PasswordSelector struct {
...
+	// OIDCCacheType
+	OIDCCacheType string `json:"oidcCacheType"`
+
+	// +kubebuilder:validaton:Optional
+	// OIDCMemCacheServers
+	OIDCMemCacheServers string `json:"oidcMemCacheServers"`
+
+	// +kubebuilder:validation:Optional
+	// KeystoneFederationIdentityProviderName
+	KeystoneFederationIdentityProviderName string `json:"KeystoneFederationIdentityProviderName,omitempty"`
```
d34dh0r53, line 258: Fix the capitalization of the JSON string name, it should be `keystoneFederationIdentityProviderName`

api/bases/keystone.openstack.org_keystoneapis.yaml
d34dh0r53: Fix this in `api/v1beta1/keystoneapi_types.go`

controllers/keystoneapi_controller.go
```diff
@@ -1204,6 +1215,44 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
+			if err != nil {
+				return err
+			}
+
+			endptConfig["EnableFederation"] = true
+			endptConfig["OIDCClaimPrefix"] = instance.Spec.OIDCFederation.OIDCClaimPrefix
+			endptConfig["OIDCResponseType"] = instance.Spec.OIDCFederation.OIDCClaimPrefix
```
d34dh0r53, line 1243: This should be instance.Spec.OIDCFederation.OIDCResponseType

api/bases/keystone.openstack.org_keystoneapis.yaml
```diff
@@ -84,6 +84,19 @@ spec:
                   files. Those get added to the service config dir in /etc/<service>
                   . TODO: -> implement'
                 type: object
+              enableFederation:
+                default: false
+                description: Enablement of Federation configuration
+                type: boolean
+              enableInsecureDebug:
```
vakwetu, line 91: enableInsecureDebug and enableKeystoneDebug  feel like settings that should be set in custom.conf.  I would imagine that in production, they would always be set to false.  This is only useful in a dev/debugging environment.  If the custom.conf setting is not working, thats a different problem.

controllers/keystoneapi_controller.go
```diff
@@ -1203,6 +1216,35 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
+		if instance.Spec.EnableFederation {
+			ospSecret, _, err := oko_secret.GetSecret(
+				ctx,
+				h,
+				instance.Spec.Secret,
+				instance.Namespace)
+			if err != nil {
+				return err
+			}
+
```
vakwetu, line 1230: Now that you have the ospsecret, you might want to validate that the contents include the KeystoneOIDCClientSecret and KeystoneOIDCCryptoPassphrase and message and error out appropriately.  Right now this will either throw an error or worse put a blank string into the config.

vakwetu (changes requested):
See comments.

vakwetu:
Maybe I missed it, but I didn't see any kuttl or functional tests .  Just the removal of a bunch of empty lines in keystoneapi_controller_test.go.  Is that coming in a separate PR?

controllers/keystoneapi_controller.go
d34dh0r53: Good call, I'll fix that.

templates/keystoneapi/config/keystone.conf
```diff
@@ -1,5 +1,7 @@
 [DEFAULT]
 use_stderr=true
+debug={{ .enableKeystoneDebug }}
+insecure_debug={{ .enableInsecureDebug }}
```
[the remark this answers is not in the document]
d34dh0r53: Yeah, I'll fix that.

d34dh0r53:
Yeah, KUTTL and functional testing is coming in a follow-on patch that I'm working on now.

api/bases/keystone.openstack.org_keystoneapis.yaml
d34dh0r53: Done

templates/keystoneapi/config/keystone.conf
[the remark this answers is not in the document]
d34dh0r53: Done

controllers/keystoneapi_controller.go
d34dh0r53: Done

api/v1beta1/keystoneapi_types.go
```diff
@@ -180,6 +180,15 @@ type KeystoneAPISpecCore struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to the TLS
 	TLS tls.API `json:"tls,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default=false
+	// Enablement of Federation configuration
+	EnableFederation bool `json:"enableFederation"`
```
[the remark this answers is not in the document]
stuggi: best to test this pr using a previous version and update to have this change

api/v1beta1/keystoneapi_types.go
```diff
@@ -180,6 +180,15 @@ type KeystoneAPISpecCore struct {
...
+
+	// +kubebuilder:validation:Optional
+	// +OIDCFederation - parameters to configure keystone for OIDC federation
+	OIDCFederation KeystoneFederationSpec `json:"oidcFederation,omitempty"`
```
stuggi, line 191: might want to do this a ptr so that it won't show up with all the empty values?

```
kind: KeystoneAPI
metadata:
  creationTimestamp: "2024-11-18T19:38:57Z"
  finalizers:
  - openstack.org/keystoneapi
  - openstack.org/keystoneservice-nova
  - openstack.org/keystoneservice-glance
  - openstack.org/keystoneservice-ceilometer
  - openstack.org/keystoneservice-swift
  - openstack.org/keystoneendpoint-glance-default-external
  - openstack.org/keystoneendpoint-glance-default-internal
  - openstack.org/keystoneendpoint-swift
  - openstack.org/keystoneservice-barbican
  - openstack.org/keystoneservice-cinderv3
  - openstack.org/keystoneendpoint-barbican-api
  - openstack.org/keystoneendpoint-cinderv3
  - openstack.org/keystoneservice-neutron
  - openstack.org/keystoneservice-placement
  - openstack.org/keystoneendpoint-neutron
  - openstack.org/keystoneendpoint-placement
  - openstack.org/keystoneendpoint-nova
  generation: 2
  name: keystone
  namespace: openstack
  ownerReferences:
  - apiVersion: core.openstack.org/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: OpenStackControlPlane
    name: openstack-galera
    uid: 54409ae3-a561-43b7-9251-603437046eb3
  resourceVersion: "70386"
  uid: fbc4eab5-72b6-485e-b9b9-6536919d4c4b
spec:
  adminProject: admin
  adminUser: admin
  containerImage: quay.io/podified-antelope-centos9/openstack-keystone@sha256:198eb79012e1b824b4092e4907f0565051f8514a7ca0d4606e9ce7a3d8439895
  databaseAccount: keystone
  databaseInstance: openstack
  enableFederation: false
  enableSecureRBAC: true
  fernetMaxActiveKeys: 5
  fernetRotationDays: 1
  memcachedInstance: memcached
  oidcFederation:
    keystoneFederationIdentityProviderName: ""
    oidcCacheType: ""
    oidcClaimDelimiter: ""
    oidcClaimPrefix: ""
    oidcClientID: ""
    oidcIntrospectionEndpoint: ""
    oidcMemCacheServers: ""
    oidcPassClaimsAs: ""
    oidcPassUserInfoAs: ""
    oidcProviderMetadataURL: ""
    oidcResponseType: ""
    oidcScope: ""
    remoteIDAttribute: ""
```

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_keystone-operator/479/pull-ci-openstack-k8s-operators-keystone-operator-main-keystone-operator-build-deploy-kuttl/1858575042430898176/artifacts/keystone-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-3d1f620ae3617fbd6371f55aa948b34e9daf31a461d3c557cf1a7c2db38d0385/namespaces/openstack/crs/keystoneapis.keystone.openstack.org/keystone.yaml

api/v1beta1/keystoneapi_types.go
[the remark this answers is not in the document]
d34dh0r53: It doesn't have to be required, I'll change it to optional

api/v1beta1/keystoneapi_types.go
stuggi: just change it to 
```
OIDCFederation *KeystoneFederationSpec `json:"oidcFederation,omitempty"`
```
probably needs some check for it bein nil where you right now just use it.

api/v1beta1/keystoneapi_types.go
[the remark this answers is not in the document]
vakwetu: do we even need this field?  If you change the federationConfig object to a pointer, you can check for null-ness

api/v1beta1/keystoneapi_types.go
vakwetu: You can see how I'm doing it with the PKCS11 settings in barbican in https://github.com/openstack-k8s-operators/barbican-operator/pull/168

d34dh0r53:
I'm running into this:
```
# github.com/openstack-k8s-operators/openstack-operator/pkg/openstack
pkg/openstack/keystone.go:115:36: cannot use instance.Spec.NodeSelector (variable of type map[string]string) as *map[string]string value in assignment
```
Building the openstack-operator that depends on this PR

vakwetu:
@d34dh0r53  have you updated to latest.  There were recently some changes by @olliewalsh for nodeSelector()

api/v1beta1/keystoneapi_types.go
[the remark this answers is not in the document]
d34dh0r53: Done

api/v1beta1/keystoneapi_types.go
d34dh0r53: Done

controllers/keystoneapi_controller.go
```diff
@@ -1190,6 +1192,16 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
 			keystone.DatabaseName,
 		),
 		"enableSecureRBAC": instance.Spec.EnableSecureRBAC,
+		"enableFederation": instance.Spec.EnableFederation,
+	}
+
+	if instance.Spec.EnableFederation {
+		federationParameters := map[string]interface{}{
+			"federationTrustedDashboard": fmt.Sprintf("https://%s-%s.%s.svc/dashboard/auth/websso/",
```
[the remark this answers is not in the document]
d34dh0r53: Yeah, this is exactly what I was looking for.  Thank you!

controllers/keystoneapi_controller.go
```diff
@@ -1205,11 +1227,49 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
+				h,
+				instance.Spec.Secret,
+				instance.Namespace)
+			if err != nil {
+				return err
+			}
+
+			OIDCClientSecret := string(ospSecret.Data[instance.Spec.PasswordSelectors.KeystoneOIDCCryptoPassphrase])
+			if OIDCClientSecret == "" {
+				return errors.New("OIDCClientSecret cannot be empty")
```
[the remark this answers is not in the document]
stuggi: we should probably also enhance L721
```
	hash, result, err := oko_secret.VerifySecret(ctx, types.NamespacedName{Name: instance.Spec.Secret, Namespace: instance.Namespace}, []string{"AdminPassword"}, helper.GetClient(), time.Second*10)
```
to make `[]string{"AdminPassword"}` conditional to check for `KeystoneOIDCCryptoPassphrase` and `KeystoneOIDCClientSecret`. this would also add them to the hash, in case the passwords change. This will not catch if the password is an empty string, but make sure an entry exist in the secret.

controllers/keystoneapi_controller.go
```diff
@@ -1197,6 +1204,21 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
+		"enableFederation": enableFederation,
+	}
+
+	if enableFederation {
+		endpointPublic, err := instance.GetEndpoint(endpoint.EndpointPublic)
+		if err != nil {
+			return err
+		}
+
+		federationParameters := map[string]interface{}{
```
stuggi, line 1216: could just add this to `templateParmeters`?

controllers/keystoneapi_controller.go
```diff
@@ -1205,11 +1227,49 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
 		endptConfig["ServerName"] = fmt.Sprintf("%s-%s.%s.svc", instance.Name, endpt.String(), instance.Namespace)
 		endptConfig["TLS"] = false // default TLS to false, and set it bellow to true if enabled
+		endptConfig["EnableFederation"] = enableFederation
 		if instance.Spec.TLS.API.Enabled(endpt) {
 			endptConfig["TLS"] = true
 			endptConfig["SSLCertificateFile"] = fmt.Sprintf("/etc/pki/tls/certs/%s.crt", endpt.String())
 			endptConfig["SSLCertificateKeyFile"] = fmt.Sprintf("/etc/pki/tls/private/%s.key", endpt.String())
 		}
+
+		if enableFederation {
```
stuggi, line 1237: I think we could just add all this to the templateParameters as they are not endpoint specific like the certs. like the httpd ProcessNumber

stuggi: or at least query the ospSecret out side the for loop

templates/keystoneapi/config/httpd.conf
```diff
@@ -51,11 +51,56 @@ CustomLog /dev/stdout proxy env=forwarded
   SSLCertificateKeyFile   "{{ $vhost.SSLCertificateKeyFile }}"
 {{- end }}
 
-  ## WSGI configuration
+ ## WSGI configuration
```
stuggi, line 54: ?

stuggi (changes requested):
I think we should have at least functional tests to validate entries get added to the httpd.conf, similar to how we check certs are correct configured:

https://github.com/openstack-k8s-operators/keystone-operator/blob/main/tests/functional/keystoneapi_controller_test.go#L1001

controllers/keystoneapi_controller.go
```diff
@@ -1205,11 +1227,49 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
+			ospSecret, _, err := oko_secret.GetSecret(
+				ctx,
+				h,
+				instance.Spec.Secret,
+				instance.Namespace)
+			if err != nil {
+				return err
+			}
+
+			OIDCClientSecret := string(ospSecret.Data[instance.Spec.PasswordSelectors.KeystoneOIDCCryptoPassphrase])
```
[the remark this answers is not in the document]
d34dh0r53: Good catch, fixed.

d34dh0r53:
Functional tests are failing with
```

**Comment by stuggi** [comment]
@d34dh0r53 have you tried to run the functional tests local? good to start with runnign them with e.g. `GINKGO_ARGS="--output-interceptor-mode=none" PROCS=1 make test` .

**Comment by millevy** [comment]
recheck

**Comment by d34dh0r53** [comment]
recheck

File: templates/keystoneapi/config/keystone.conf [hunk 23]
```diff
@@ -11,6 +11,17 @@ enabled=true
...
+{{if .enableFederation }}
+[federation]
+trusted_dashboard={{ .federationTrustedDashboard }}
+
+[openid]
+remote_id_attribute={{ .federationRemoteIDAttribute }}
+
+[auth]
+methods = password,token,oauth1,mapped,application_credential,openid
+{{ end }}
```
stuggi, line 23: do we need to add this using an api parameter? we do not want to add tht parameters via the service operators api. a user could use the native interface with customServiceConfig to customize it.

api/v1beta1/keystoneapi_types.go
```diff
@@ -199,6 +203,79 @@ type PasswordSelector struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="HTTP_OIDC_ISS"
+	// RemoteIDAttribute
+	RemoteIDAttribute string `json:"remoteIDAttribute"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	// KeystoneFederationIdentityProviderName
+	KeystoneFederationIdentityProviderName string `json:"keystoneFederationIdentityProviderName"`
 }
```
stuggi, line 279: we probably should think of a generic config customizion interface for httpd, like we have with customServiceConfig to not require most of these parameters. but right now we do not have it.

stuggi:
just fyi, I started today to look into if we can have a generic interface to customize/extend the httpd service vhost config and if we could use it for this scenario. I have an early version for a static extension. For supporting a template like for this I need to do some more work. Should have something maybe by mid of next week which we could test. just want to let you know and maybe we wait landing this until then?

stuggi:
> just fyi, I started today to look into if we can have a generic interface to customize/extend the httpd service vhost config and if we could use it for this scenario. I have an early version for a static extension. For supporting a template like for this I need to do some more work. Should have something maybe by mid of next week which we could test. just want to let you know and maybe we wait landing this until then?

@d34dh0r53 @vakwetu please have a look at https://github.com/openstack-k8s-operators/keystone-operator/pull/525 which I'll bring up to the wider team as a generic way to customize and template custom configs and also being able to customize the httpd vhost config.

d34dh0r53:
With the addition of PR #525 this patch is no longer needed.  The httpd customization can be performed using a secret and the changes needed for keystone.conf are handled in `customServiceConfig`.

PR #480, opened 2024-10-10
Make sure to log error in findObjectsForSrc()

stuggi:
It is hidden right now if there is an error in configured named fields and index. Lets log the error and return the current state of requests.

With this the findObjectsForSrc() will exit with an hidden error like:

```
"error": "Index with name field:.spec.ksmTls.caBundleSecretName does not exist"
```


abays (approved):
/lgtm

PR #489, opened 2024-10-21
Fernet rotation kuttl tests

xek:
Jira: [OSPRH-4376](https://issues.redhat.com//browse/OSPRH-4376)

### DIFF
--- a/tests/kuttl/common/scripts/validate_test_token.sh
+++ b/tests/kuttl/common/scripts/validate_test_token.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+set -euxo pipefail
+
+seconds=1
+while [ $seconds -le 30 ]; do
+    rotatedat=$(oc get secret keystone -n $NAMESPACE -o jsonpath="{.metadata.annotations['keystone\.openstack\.org/rotatedat']}")
+    if [ $rotatedat != "2009-11-10T23:00:00Z" ]; then
+        break
+    fi
+    sleep 1
+    seconds=$(( $seconds + 1 ))
+done
+
+sleep 20 # make sure a rollout started
+
+oc rollout status deployment/keystone -n $NAMESPACE
+
+export OS_TOKEN=$(cat /tmp/temporary_test_token)
+
+alias openstack="oc exec -tn $NAMESPACE openstackclient -- env -u OS_CLOUD - OS_AUTH_URL=http://keystone-public.keystone-kuttl-tests.svc:5000 OS_AUTH_TYPE=token OS_TOKEN=$OS_TOKEN openstack"
+
+if openstack endpoint list 2>&1 | grep "Failed to validate token"; then
+    exit 1
+else
+    exit 0
+fi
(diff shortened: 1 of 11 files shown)


xek:
Get \"https://quay.rdoproject.org/v2/\": dial tcp 38.129.56.158:443: i/o timeout"

tests/kuttl/tests/fernet_rotation/02-assert.yaml
```diff
@@ -0,0 +1,6 @@
+# validate that token saved in previous step works
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: $KEYSTONE_KUTTL_DIR/../common/scripts/validate_test_token.sh
```
stuggi, line 6: starting with kuttl 0.17.0 should be relative to the test dir, so `../../common/scripts/validate_test_token.sh`. otherwise you could not run it manually from the base dir

tests/kuttl/tests/fernet_rotation/01-assert.yaml
```diff
@@ -0,0 +1,5 @@
+# create a test token, save it in /tmp/
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: $KEYSTONE_KUTTL_DIR/../common/scripts/create_test_token.sh
```
stuggi, line 5: starting with kuttl 0.17.0 should be relative to the test dir, so `../../common/scripts/validate_test_token.sh`. otherwise you could not run it manually from the base dir

tests/kuttl/common/scripts/validate_test_token.sh
```diff
@@ -0,0 +1,22 @@
...
+seconds=1
+while [ $seconds -le 30 ]; do
+    rotatedat=$(oc get secret keystone -n $NAMESPACE -o jsonpath="{.metadata.annotations['keystone\.openstack\.org/rotatedat']}")
+    if [ $rotatedat != "2009-11-10T23:00:00Z" ]; then
+        break
+    fi
+    sleep 1
+    seconds=$(( $seconds + 1 ))
+done
+
```
stuggi, line 13: we have to wait for the keystoneapi to be restarted. otherwise we just test against the old deployment. when the annotation on the secret got updated it does not mean the deployment pod already got updated.
It could be a stupid sleep for maybe 20 or 30s, or probably better to exec into the keystone pod and check if it has the rotated keys on place.

stuggi (changes requested):

tests/kuttl/common/scripts/validate_test_token.sh
xek: You're right. I'll opt for the sleep, since the mounted secret data can change without restarting the pod.

xek: I'll sleep for 20s and then do` oc rollout status deployment keystone -n openstack`

stuggi (approved):
/lgtm

PR #490, opened 2024-10-22
[test] enhance functional tests to check all fernet keys

stuggi:
Checks all fernet keys are rotated correct:
* old idx 0 > new 4
* old idx   > new idx-1
* old idx 1 must be gone and not match new idx 0


xek (approved):
/lgtm Thanks!

PR #492, opened 2024-10-24
Add tests for custom number of fernet keys

afaranha:


tests/functional/keystoneapi_controller_test.go
```diff
@@ -1109,6 +1110,112 @@ var _ = Describe("Keystone controller", func() {
 		})
 	})
 
+	When("When the fernet keys are created with fernetMaxActiveKeys as 3", func() {
```
afaranha, line 1113: Both this test and the one below are pretty much similar, is there a way to reuse it?
(While waiting for an answer I'll try to figure it out, but it may take some time)

afaranha:
This is a draft because I still need to add tests changing the fernetMaxActiveKeys and having it reflected on the environment.

tests/functional/keystoneapi_controller_test.go
```diff
@@ -1109,6 +1110,112 @@ var _ = Describe("Keystone controller", func() {
...
+			Eventually(func(g Gomega) {
+				numberFernetKeys := 0
+				for k, _ := range secret.Data {
+					if strings.HasPrefix(k, "FernetKeys") {
+						numberFernetKeys++
+					}
+				}
+
+				Expect(numberFernetKeys).Should(BeNumerically("==", 3))
+				for i := 0; i < 2; i++ {
```
stuggi, line 1159: `i < 3` or `i <= 2`

tests/functional/keystoneapi_controller_test.go
```diff
@@ -1109,6 +1110,112 @@ var _ = Describe("Keystone controller", func() {
...
+					g.Expect(secret.Data["FernetKeys"+strconv.Itoa(i)]).NotTo(BeNil())
```
stuggi, line 1160: this won't be nil when the key exists. the value could be the empty string, so probably we want instead of `BeNil()` check for `BeZero()`

tests/functional/keystoneapi_controller_test.go
stuggi: instead of the separate test bellow, maybe it would be better to test that when updating the keystone parameter from 3 -> 6 we see that the config and the secret got additional keys? Like have a nested when update keys from 3->6 it ...

tests/functional/keystoneapi_controller_test.go
```diff
@@ -1109,6 +1110,112 @@ var _ = Describe("Keystone controller", func() {
...
+				}
```
stuggi, line 1161: Instead of the looping, could  do this check for the expected Fernet keys 0-2 and not to have a key3

```
                                Expect(maps.Keys(secret.Data)).To(ContainElements("FernetKeys0", "FernetKeys1", "FernetKeys2"))
				Expect(maps.Keys(secret.Data)).ToNot(ContainElement("FernetKeys3"))

				for i := 0; i < 3; i++ {
					g.Expect(secret.Data["FernetKeys"+strconv.Itoa(i)]).NotTo(BeZero())
				}
```

templates/keystoneapi/config/keystone.conf
```diff
@@ -22,7 +22,7 @@ enforce_scope = {{ .enableSecureRBAC }}
 
 [fernet_tokens]
 key_repository=/etc/keystone/fernet-keys
-max_active_keys=2
+max_active_keys={{ .fernetMaxActiveKeys }}
```
xek, line 25: This setting is only used for Fernet key rotation [1], which we implemented in go. I guess we can keep your change, to be consistent...

[1] https://github.com/openstack/keystone/blob/6433f1ae1c094e58216360d191a21d956a80f658/keystone/common/fernet_utils.py#L261

tests/functional/keystoneapi_controller_test.go
afaranha: The problem is that I still don't know how to simulate a change on the spec.
I start with this one: DeferCleanup(th.DeleteInstance, CreateKeystoneAPI(keystoneAPIName, GetKeystoneAPISpec(3)))
But even if I call this again with a different value on the It(), it doesn't seem to work.

stuggi: you could do it like this

```
	When("When the fernet keys are created", func() {
		BeforeEach(func() {
			DeferCleanup(
				k8sClient.Delete, ctx, CreateKeystoneMessageBusSecret(namespace, "rabbitmq-secret"))
			// with fernetMaxActiveKeys as 3
			DeferCleanup(th.DeleteInstance, CreateKeystoneAPI(keystoneAPIName, GetKeystoneAPISpec(3)))
			DeferCleanup(
				k8sClient.Delete, ctx, CreateKeystoneAPISecret(namespace, SecretName))
			DeferCleanup(infra.DeleteMemcached, infra.CreateMemcached(namespace, "memcached", memcachedSpec))
			DeferCleanup(
				mariadb.DeleteDBService,
				mariadb.CreateDBService(
					namespace,
					GetKeystoneAPI(keystoneAPIName).Spec.DatabaseInstance,
					corev1.ServiceSpec{
						Ports: []corev1.ServicePort{{Port: 3306}},
					},
				),
			)
			mariadb.SimulateMariaDBAccountCompleted(keystoneAccountName)
			mariadb.SimulateMariaDBDatabaseCompleted(keystoneDatabaseName)
			infra.SimulateTransportURLReady(types.NamespacedName{
				Name:      fmt.Sprintf("%s-keystone-transport", keystoneAPIName.Name),
				Namespace: namespace,
			})
			infra.SimulateMemcachedReady(types.NamespacedName{
				Name:      "memcached",
				Namespace: namespace,
			})
			th.SimulateJobSuccess(dbSyncJobName)
			th.SimulateJobSuccess(bootstrapJobName)
			th.SimulateDeploymentReplicaReady(deploymentName)
		})

		It("creates 3 keys", func() {
			secret := th.GetSecret(types.NamespacedName{Namespace: keystoneAPIName.Namespace, Name: "keystone"})
			Expect(secret).ToNot(BeNil())

			Eventually(func(g Gomega) {
				Expect(maps.Keys(secret.Data)).To(ContainElements("FernetKeys0", "FernetKeys1", "FernetKeys2"))
				Expect(maps.Keys(secret.Data)).ToNot(ContainElement("FernetKeys3"))

				for i := 0; i < 3; i++ {
					g.Expect(secret.Data["FernetKeys"+strconv.Itoa(i)]).NotTo(BeZero())
				}
			}, timeout, interval).Should(Succeed())
		})

		When("When the fernet keys get updated to 6", func() {
			BeforeEach(func() {
				keystone := GetKeystoneAPI(keystoneAPIName)
				Expect(*keystone.Spec.FernetMaxActiveKeys).To(Equal(int32(3)))

				keystone.Spec.FernetMaxActiveKeys = ptr.To(int32(6))

				_, err := controllerutil.CreateOrPatch(
					th.Ctx, th.K8sClient, keystone, func() error { return nil })
				Expect(err).ToNot(HaveOccurred())
			})
			It("creates 6 keys", func() {
				secret := th.GetSecret(types.NamespacedName{Namespace: keystoneAPIName.Namespace, Name: "keystone"})
				Expect(secret).ToNot(BeNil())

				Eventually(func(g Gomega) {
					Expect(maps.Keys(secret.Data)).To(ContainElements("FernetKeys0", "FernetKeys1", "FernetKeys2", "FernetKeys3", "FernetKeys4", "FernetKeys5"))
					Expect(maps.Keys(secret.Data)).ToNot(ContainElement("FernetKeys6"))

					for i := 0; i < 6; i++ {
						g.Expect(secret.Data["FernetKeys"+strconv.Itoa(i)]).NotTo(BeZero())
					}
				}, timeout, interval).Should(Succeed())
			})
		})
	})
```

when you run this it fails with

```
  [FAILED] Expected
      <[]string | len:5, cap:5>: ["FernetKeys2", "CredentialKeys0", "CredentialKeys1", "FernetKeys0", "FernetKeys1"]
  to contain elements
      <[]string | len:6, cap:6>: ["FernetKeys0", "FernetKeys1", "FernetKeys2", "FernetKeys3", "FernetKeys4", "FernetKeys5"]
  the missing elements were
      <[]string | len:3, cap:3>: ["FernetKeys3", "FernetKeys4", "FernetKeys5"]
  In [It] at: /home/mschuppe/src/github.com/openstack-k8s-operators/keystone-operator/tests/functional/keystoneapi_controller_test.go:1178 @ 10/25/24 12:34:45.138
```

so I think there might be a bug with the logic which adds additional keys?

afaranha: Thanks for the tip on updating the spec.
Regarding the test part:
```
				Expect(maps.Keys(secret.Data)).To(ContainElements("FernetKeys0", "FernetKeys1", "FernetKeys2"))
				Expect(maps.Keys(secret.Data)).ToNot(ContainElement("FernetKeys3"))
```
I prefer to make sure there's no leftovers, for example, if we decrease from 6 to 3 keys, if somehow we ended up in this scenario: FernetKeys0, 1, 2, 5; I wanna make sure ALL the 'FernetKeys%d' are what we expect, so I`m counting how many FK we have and that they are all what we expect. Other way to do this (That I know of) would be to get each FK and see if they are valid; to check that they are valid would be to iterate over a range of fernetMaxKeys, create a string like 'FernetKeys%d' and add it to a list.
I preferred just to count the FernetKeys keys on the secret.

tests/functional/keystoneapi_controller_test.go
```diff
@@ -1109,6 +1111,118 @@ var _ = Describe("Keystone controller", func() {
...
+			th.SimulateDeploymentReplicaReady(deploymentName)
+		})
+
+		It("creates 6 keys", func() {
+			secret := th.GetSecret(types.NamespacedName{Namespace: keystoneAPIName.Namespace, Name: "keystone"})
+			Expect(secret).ToNot(BeNil())
+
+			keystone := GetKeystoneAPI(keystoneAPIName)
+			keystone.Spec.FernetMaxActiveKeys = ptr.To(int32(6))
+			_, err := controllerutil.CreateOrPatch(
```
afaranha, line 1206: This is not raising an error but the test doesn't work, still keeps 6 keys, maybe that's a bug. I`ll investigate it.

tests/functional/keystoneapi_controller_test.go
stuggi: I am not saying you should use from above, but you could still do the same with that, like what I did that a key must not exist. I prefer that as it is easy to read. e.g. when you go from 6 to 3, you:
* expect these
Expect(maps.Keys(secret.Data)).To(ContainElements("FernetKeys0", "FernetKeys1", "FernetKeys2"))
* you expect that those are gone
 Expect(maps.Keys(secret.Data)).ToNot(ContainElements("FernetKeys3", "FernetKeys4", "FernetKeys5"))

What you probably want to enhance, e.g. when you go from 3 to 6 that the last key is the expected one. iiuc in this scenario, the old key[2] should become the key[5], right?

tests/functional/keystoneapi_controller_test.go
stuggi: why should this raise an error, with this you update the keystoneapi CR to have the value changed from 3 to 6

afaranha: I`m not expecting an error here, the next line is to enforce there was no issue during the update of values:
`Expect(err).ToNot(HaveOccurred())`

stuggi: yes correct, I was reading your first message here that you'd expect an error. sorry misunderstood that.

afaranha:
Closing this PR because I messed up the rebase.
Opened this one instead: https://github.com/openstack-k8s-operators/keystone-operator/pull/497

PR #494, opened 2024-10-29
Switch to use new CreateServiceReadyCondition

stuggi:
The current ExposeServiceReadyCondition can be confusing since it just reflects if the k8s service got created, but not the route to expose the public endpoint. This changes to a new CreateServiceReadyCondition to prevent this.

Jira: OSPRH-4164

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/572


abays (approved):
/lgtm

PR #500, opened 2024-11-13
[WIP] Make process number configurable in KeystoneAPI

Deydra71:
The process number interval is yet TBD. 

Jira: [OSPRH-10363](https://issues.redhat.com/browse/OSPRH-10363)


api/bases/keystone.openstack.org_keystoneapis.yaml
```diff
@@ -310,6 +310,14 @@ spec:
                 description: PreserveJobs - do not delete jobs after they finished
                   e.g. to check logs
                 type: boolean
+              processNumber:
+                default: 3
+                description: Processumber - Number of processes running in keystone
+                  API
+                format: int32
+                maximum: 10
```
vakwetu, line 318: not sure about setting a max here.  @d34dh0r53 @dmendiza @xek   - is there a reasonable setting here or should we just leave it blank?

Deydra71: I deleted the maximum limit for now for testing.

d34dh0r53: I also don't think a maximum is needed, it's easy enough to add should this become a problem.

vakwetu (approved):

stuggi:
This is probably something we want to be able to do across multiple service operators api deployments, but I am not sure if adding it as a top parameter for the deployment is good. if there will be more parameters we need to be able to customize related to httpd its probably better to have them in a sub struct. And this needs to be the same for all operators. @dprince @olliewalsh @abays thoughts?

maybe we should have a generic crd for httpd which allows us to extend for future customization.

vakwetu:
@d34dh0r53  ^^ just for awareness of the above.  @d34dh0r53  is working on changes needed for federation.

olliewalsh:
> This is probably something we want to be able to do across multiple service operators api deployments, but I am not sure if adding it as a top parameter for the deployment is good. if there will be more parameters we need to be able to customize related to httpd its probably better to have them in a sub struct. And this needs to be the same for all operators. @dprince @olliewalsh @abays thoughts?
> 
> maybe we should have a generic crd for httpd which allows us to extend for future customization.

+1 for a sub struct. Also could potentially switch from apache/wsgi to $something_else in future

vakwetu:
I'm not opposed to a substruct of some sort.  That said, there are types of apache config which are much more extensive - see https://github.com/openstack-k8s-operators/keystone-operator/pull/479  Does the approach there change at all?

api/bases/keystone.openstack.org_keystoneapis.yaml
```diff
@@ -310,6 +310,13 @@ spec:
                 description: PreserveJobs - do not delete jobs after they finished
                   e.g. to check logs
                 type: boolean
+              processNumber:
+                default: 3
+                description: Processumber - Number of processes running in keystone
```
dprince, line 315: Processumber -> Processnumber

stuggi:
> > This is probably something we want to be able to do across multiple service operators api deployments, but I am not sure if adding it as a top parameter for the deployment is good. if there will be more parameters we need to be able to customize related to httpd its probably better to have them in a sub struct. And this needs to be the same for all operators. @dprince @olliewalsh @abays thoughts?
> > maybe we should have a generic crd for httpd which allows us to extend for future customization.
> 
> +1 for a sub struct. Also could potentially switch from apache/wsgi to $something_else in future

I was thinking about adding some subsection for `httpdTunings` or `httpdCustomization`. if there are more settings, we could just add them there, instead of having multiple top level ones. 
```
spec:
  httpdCustomization:
    processNumber:
```

don't think it affects https://github.com/openstack-k8s-operators/keystone-operator/pull/479 as it is a separate feature which are not a generic httpd settings.

stuggi (commented):
looks good to me, @olliewalsh @dprince @abays for you too?

api/v1beta1/keystoneapi_types.go
```diff
@@ -197,6 +201,15 @@ type PasswordSelector struct {
 	Admin string `json:"admin"`
 }
 
+// HttpdCustomization - customize the httpd service
+type HttpdCustomization struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=3
+	// +kubebuilder:validation:Minimum=1
```
stuggi, line 208: as discussed on slack, this needs to be tested for updating the controller/CRD on an already deployed env

olliewalsh:
looks good to me, if updating works

stuggi (approved):
/lgtm

PR #501, opened 2024-11-14
kuttl: switch reports to JUnitXML instead of JSON

jparoly:
JUnitXML is the de-facto standard format for test reporting and widely supported by all tools which post-process test results.


tosky:
I think this one can be closed, as it was superseded by https://github.com/openstack-k8s-operators/keystone-operator/pull/547

PR #503, opened 2024-11-19
Add test to rotate keys until an old one invalidades

afaranha:
Jira: [OSPRH-9674](https://issues.redhat.com//browse/OSPRH-9674)


tests/kuttl/tests/fernet_rotation/03-assert.yaml
```diff
@@ -0,0 +1,6 @@
+# validate that token saved in previous step works
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: ../../common/scripts/test_invalid_token.sh
```
Deydra71, line 6: The file doesn't exit, should it be `validate_test_token.sh` instead?

tests/kuttl/tests/fernet_rotation/03-rotate_keys.yaml
```diff
@@ -0,0 +1,153 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: keystone
+  annotations:
+    keystone.openstack.org/rotatedat: "2009-11-10T23:00:00Z"
+commands:
+  - script: |
+      #!/bin/sh
```
Deydra71, line 9: This script is repeated after every secret. Maybe create a script with this content and just call the script to keep it more clear and avoid redundancy?

tests/kuttl/tests/fernet_rotation/03-assert.yaml
afaranha: Thanks, it didn't go with `git add .`
Just submitted a new patch with that instead.
That's a different script, while `validate_test_token.sh` expects the token to work, `test_invalid_token.sh` does the same but the token should fail.

tests/kuttl/tests/fernet_rotation/03-rotate_keys.yaml
afaranha: Yea, agreed, I was waiting for it to pass the tests to refactor, but that's minor change, I'll add that and `validate_test_token.sh` and `test_invalid_token.sh` since they are both basically the same.

tests/kuttl/tests/fernet_rotation/03-assert.yaml
Deydra71: Oh okay, thanks for clarifying that!

Deydra71:
Updates:

Replaced `oc patch` with `oc apply` to handle the entire resource replacement reliably. `oc patch` often resulted in inconsistent updates or partial changes in keystone secret, which resulted in incomplete rotation triggers (secret annotation was updated, but fernet keys weren't rotated... but this was not happening 100% of times.)
`oc apply `ensures the full configuration is applied cleanly.

Increased the timeout duration for rollout status checks to accommodate slower deployments observed during rotations. The 100 seconds might not be necessary for CI, but in my environment it was the first value that prevented premature failures.

TODO 1: Check the `validate_test_token.sh` script, the grep is probably not working correctly (use the updated one from `test_invalid_token.sh` script.

TODO 2: Resolve the script `hack/configure_local_webhook.sh`  added by mistake, it's probably a remnant from rebasing...

Deydra71:

timeout

tests/kuttl/common/scripts/rotate_token.sh
```diff
@@ -0,0 +1,42 @@
...
+kind: Secret
+metadata:
+    name: keystone
+    namespace: keystone-kuttl-tests
+    annotations:
+        keystone.openstack.org/rotatedat: "2009-11-10T23:00:00Z"
+EOF
+}
+
+for rotation in {1..6}; do
```
xek, line 18: We should do exactly 3 rotations, right? The default is to store 5 keys, and we already do 1 rotation in a previous step. So discounting the staging key, we are left with 3

xek (changes requested):

tests/kuttl/common/scripts/rotate_token.sh
xek: We could try flushing the memcached cache by running
`
oc rsh -n barbican-kuttl-tests memcached-0 bash -c 'echo flush_all > /dev/tcp/localhost/11211'
`
I don't think anything else should be caching the token...

Deydra71: The command did not do anything in my env (using correct namespace). I check the files where the keys are saved within a pod, and they are rotated according to the secret, so that looks good.

@stuggi  iirc you helped to implement the fernet key rotation, do you have any idea what can be happening?

xek (approved):

tests/kuttl/common/scripts/rotate_token.sh
xek: I'm approving this for now, since these tests is a step forward, but I'm also adding a spike, to figure this out.

Deydra71: Yes, thanks for the spike to not let this slip away.

stuggi (approved):
/lgtm

PR #510, opened 2024-11-25
[renovate]Enable bumps on 18.0-fr1

gibizer:
This enables renovate run on 18.0-fr1 branch. Please note that we have a
special config for 18.0-fr1 enabled in openstack-k8s-operators#509

Related: [OSPRH-8355](https://issues.redhat.com//browse/OSPRH-8355)


gibizer:
f1 is basically end of life upstream now that we have fr2, so this is not needed any more

PR #511, opened 2024-11-27
Configure dogpile.cache to deal with memcached pods failures

lmiccini:
Whenever one of the mecached pods disappears, because of a rolling restart during a minor update or as result of a failure, APIs can take a long time to detect that the pod went away and keep trying to reconnect.

From a quick round of tests we saw downtimes up to ~150s.

By enabling the retry_client and limiting the number of retries the behavior seems much more acceptable.

Similarly, when TLS is not in use, we may want to set a lower value for memcache_dead_retry so to eventually reconnect to a new pod (having the same dns name but different ip) much faster.

Jira: https://issues.redhat.com/browse/OSPRH-11935

### DIFF
--- a/templates/keystoneapi/config/keystone.conf
+++ b/templates/keystoneapi/config/keystone.conf
@@ -5,9 +5,13 @@ use_stderr=true
 {{if .MemcachedTLS}}
 backend = dogpile.cache.pymemcache
 memcache_servers={{ .MemcachedServers }}
+enable_retry_client = true
+retry_attempts = 2
+retry_delay = 0
 {{else}}
 backend = dogpile.cache.memcached
 memcache_servers={{ .MemcachedServersWithInet }}
+memcache_dead_retry = 10
 {{end}}
 enabled=true
 tls_enabled={{ .MemcachedTLS }}


templates/keystoneapi/config/keystone.conf
```diff
@@ -4,8 +4,12 @@ use_stderr=true
 [cache]
 {{if .memcachedTLS}}
 backend = dogpile.cache.pymemcache
+enable_retry_client = true
+retry_attempts = 2
+retry_delay = 0
 {{else}}
 backend = dogpile.cache.memcached
+memcache_dead_retry = 10
```
bogdando, line 12: from puppet tripleo I can see we used to have it a 60 for all services

lmiccini: We can't really compare vs tripleo unfortunately. There we have static ips (controllers ips do not change) while here every time we get a new pod the ip changes, so we need to react faster imho.

templates/keystoneapi/config/keystone.conf
```diff
@@ -5,9 +5,13 @@ use_stderr=true
 {{if .memcachedTLS}}
 backend = dogpile.cache.pymemcache
 memcache_servers={{ .memcachedServers }}
+enable_retry_client = true
+retry_attempts = 2
+retry_delay = 0
```
xek, line 10: I think there should be at least _some_ delay between the retries, however small.

lmiccini: this is just the default value as set in https://docs.openstack.org/keystone/latest/configuration/samples/keystone-conf.html , plan is to surface this in the docs so that users can set it to the desired value.

lmiccini:
@xek @vakwetu do you guys think we can merge this? I will follow up re: documentation in Jira.

PR #518, opened 2024-12-11
Fix default cache config for no tls

karelyatin:
When env is deployed with tls disabled, cache config was wrong as "dogpile.cache.memcached" backend requires inet[6] prefixes for memcache_servers setting while it was set without the prefix.

With IPv4 issue not observed as "inet" is default prefix, this patch fixes it by setting [cache]/memcache_servers based on tls config.

Resolves: [OSPRH-12222](https://issues.redhat.com//browse/OSPRH-12222)
Related-Issue: [OSPRH-12221](https://issues.redhat.com//browse/OSPRH-12221)


stuggi (approved):
/lgtm

xek (approved):
Thanks!

PR #520, opened 2024-12-12
crd-schema-check: allow new CRDs

stuggi:
If the 'git show BASE_REF' fails assume it is a net new CRD and should be fine.


abays (approved):
/lgtm

PR #526, opened 2025-01-23
Modified CreateKeystoneAPI to take an additional parameter spec

mrkisaolamb:
This change will fix all func tests in operators without any additional changes required.

Currently tests are failing in renovate bump patches https://github.com/openstack-k8s-operators/nova-operator/pull/918 . Problem is with defaulting APITimeout, webhook is failing when spec is empty.  Unfortunattly other option with using Unstruct is required much more changes inside the function and also in other operators test suite to add keystone defaulting

### DIFF
--- a/api/test/helpers/crd.go
+++ b/api/test/helpers/crd.go
@@ -65,7 +65,11 @@ func (th *TestHelper) CreateKeystoneAPI(namespace string) types.NamespacedName {
 			Name:      "keystone-" + uuid.New().String(),
 			Namespace: namespace,
 		},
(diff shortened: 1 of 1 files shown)


gibizer (commented):
Make sense. Just prepare to update all the operators using this function in their test as the signature changes.

gibizer:
Short trim I'm OK with this to stop the bleeding. Longer term we should be careful duplicating default values between the kubebuilder def and the envtest helper as it can easily lead to divergence between the two value and we will end up testing with what is not the default any more.

PR #528, opened 2025-01-29
Fix missing APITimeout in CreateKeystoneAPIWithFixture

gthiemonge:
CreateKeystoneAPIWtithFixture doesn't set the APITimeout The other operators which use this function see an error:

  KeystoneAPI.keystone.openstack.org ... is invalid: spec.apiTimeout:
  Invalid value: 0: spec.apiTimeout in body should be greater than or
  equal to 10

Note: this function is not called by keystone-operator but it is used by the octavia-operator.


gibizer (approved):
This is aligned with the fix in https://github.com/openstack-k8s-operators/keystone-operator/pull/526

PR #530, opened 2025-02-04
Improve error handling and gosec measures

Deydra71:
To follow best practices for error and exception handling in Go and to prevent regressions, err113 and gosec checks in golang-ci-linter are now enabled.

JIra: [OSPRH-13205](https://issues.redhat.com/browse/OSPRH-13205)


.golangci.yaml
```diff
@@ -7,5 +7,12 @@ linters:
...
     - gofmt
     - govet
+    - gosec
+    - errname
+    - err113
 run:
   timeout: 5m
+
+issues:
+  exclude:
```
Deydra71, line 17: Not sure if this is the best way to suppress false positive warnings.

pkg/keystone/fernet.go
```diff
@@ -16,15 +16,18 @@ limitations under the License.
...
-// GenerateFernetKey -
+// GenerateFernetKey - returns a base64-encoded, 32-byte key using cryptographically secure random generation
 func GenerateFernetKey() string {
 	data := make([]byte, 32)
-	for i := 0; i < 32; i++ {
-		data[i] = byte(rand.Intn(10))
+
+	_, err := rand.Read(data)
+	if err != nil {
+		panic(err)
```
mauricioharley, line 29: We should not panic whatsoever.  This dumps a reasonable amount of data absolutely out of our control. We should rather log the error in a safe way.

.golangci.yaml
Deydra71: Other option is to set `// #nosec G101` in all lines where code affected with hardcoded credentials. I have not figure out way to exclude only e.g. `tests` dir

mauricioharley:
The inspiring PR (https://github.com/openstack-k8s-operators/nova-operator/pull/916) is still open. Changes on it might impact what we are doing here. Let's hold it for a while.

.golangci.yaml
xek: I didn't try it, but maybe something like:

```
exclude-rules:
  - path: tests
    text: "G101: Potential hardcoded credentials"
    linters:
      - gosec
```

xek (approved):

.golangci.yaml
Deydra71: Yup, that was it, thanks! So, now we have to add `// #nosec G101` to all lines where code is affected with hardcoded credentials, while test dir is completely excluded.

xek (approved):
/lgtm

vakwetu (approved):

pkg/keystone/fernet.go
```diff
@@ -16,15 +16,22 @@ limitations under the License.
...
 	"encoding/base64"
-	"math/rand"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-// GenerateFernetKey -
+// GenerateFernetKey - returns a base64-encoded, 32-byte key using cryptographically secure random generation
 func GenerateFernetKey() string {
+	logger := log.Log.WithName("GenerateFernetKey")
```
stuggi, line 27: maybe just pass the logger `r.GetLogger(ctx)` to GenerateFernetKey() instead of creating each time a new one?

api/v1beta1/conditions.go
```diff
@@ -111,4 +111,7 @@ const (
 
 	// KeystoneServiceOSUserReadyErrorMessage
 	KeystoneServiceOSUserReadyErrorMessage = "Keystone Service user error occured %s"
+
+	// KeystoneNetworkAttachmentsReadyErrorMessage
+	KeystoneNetworkAttachmentsReadyErrorMessage = "NetworkAttachments error occurred; not all pods have interfaces with IPs as configured in NetworkAttachments: %s"
```
stuggi, line 116: lets use the common one Kamil added to lib-common https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/condition/conditions.go#L272

stuggi (approved):
/lgtm

Deydra71:
Note:
https://github.com/openstack-k8s-operators/openstack-operator/pull/1306 needs to be merged first, until then kuttl will be failing

PR #532, opened 2025-02-10
Update topology reference from lib-common to infra-operator

fmount:
This patch updates the topology common function references to point to infra-operator, which is now the only source of truth when we import Topology related functions.


stuggi (approved):
/lgtm

PR #535, opened 2025-02-17
Use quay hosted kube-rbac-proxy image

stuggi:
The image gcr.io/kubebuilder/kube-rbac-proxy is deprecated and will become unavailable. Sometime from early 2025 the GCR will go away.
The plan is to update the project to use the new WithAuthenticationAndAuthorization from kubebuilder. Until we are able to do that, use our own hosted kube-rbac-proxy image.

Jira: OSPRH-14114


abays (approved):
/lgtm

PR #547, opened 2025-03-10
More zuul jobs, and generate JUnitXML reports for KUTTL

tosky:
Add KUTTL and tempest zuul jobs (meant to replace/complement the prow ones in the long term)

KUTTL: generate JUnitXML reports
- switch the type to XML (the accepted value is 'xml' lowercase, not uppercase);
- tune the name a bit to highlight it is a report;
- add the new reportGranularity parameter which is supported by kuttl 0.20.0 and will restore the pre-1.17 JUnitXML format (granularity by test case, instead of by step).
- remove the URL of kuttl, so that it is not tied to a specific release.

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -6,16 +6,14 @@
 # ASSUMPTIONS:
 #
 # 1. Latest version of kuttl is installed at /usr/local/bin/kubectl-kuttl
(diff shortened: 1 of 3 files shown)


d34dh0r53 (approved):
/lgtm

PR #552, opened 2025-03-25
Poll deployment on update

stuggi:
When a deployment gets updated (config, image or changed spec) and the rollout results in the deployment to fail it should be reflected in the Deployment condition. Currently if there is an issue on the rollout of the deployment and the rollout/update results in a ProgressDeadlineExceed it is not reflected in the service operators conditions.
Instead because the old config replicas are still up and healthy, the service is still functional with the old specs, the Deployment conditions shows ready, but the rollout is stuck and failed to deploy what was requested.

Example:
~~~
NAME       NETWORKATTACHMENTS   STATUS   MESSAGE
keystone                        True     Setup complete
keystone                        True     Setup complete <<<< depl complete

>> add broken httpd config, but could be a broken image, or someething else which makes the pod fail to start
keystone                        Unknown   Service config create not started
keystone                        Unknown   Setup started
keystone                        False     rollout status: 1/4 replicas updated
>>> replacement pod is failing
keystone                        False     rollout status: 1/4 replicas updated
...
keystone                        False     keystone ProgressDeadlineExceeded - ReplicaSet "keystone-74db779db5" has timed out progressing.
>>> DeadlineExceeded reached (default 10min)
~~~

As a result the keystoneAPI is not ready, because keystone is not ready for the new config requested/rollout failed. Other services which rely on the keystoneapi will stop at:

~~~
$ oc get neutronapi -n openstack
NAME      NETWORKATTACHMENTS                                                            STATUS   MESSAGE
neutron   {"openstack/internalapi":["172.17.0.33"],"ovn-kubernetes":["10.217.0.169"]}   False    KeystoneAPI not yet ready
~~~

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/613

Jira: [OSPRH-14472](https://issues.redhat.com//browse/OSPRH-14472)

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -1128,6 +1128,28 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
 			condition.DeploymentReadyRunningMessage))
 		return ctrlResult, nil
 	}
+
+	// Check the rollout status and reflect condition in case of
+	// an ongoing deployment update
+	if !depl.RolloutComplete() && depl.GetRolloutStatus() != nil {
+		switch *depl.GetRolloutStatus() {
+		case deployment.DeploymentPollProgressing:
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.DeploymentReadyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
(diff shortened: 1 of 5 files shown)


stuggi:
closing in favor of a less invasive solution, at least atm

PR #554, opened 2025-03-27
Add config for Domain-specific LDAP Backends

xek:
This mounts the provided secret as a volume into
/etc/keystone/domains.


fmount:
@xek this use case would be solved by adding `ExtraMounts` interface to keystone (see https://github.com/openstack-k8s-operators/keystone-operator/pull/559).

PR #555, opened 2025-03-27
Run keystone services using keystone user

fmount:
This patch represents an improvement of the existing code to make sure we run keystone services using the keystone user instead of root.

Jira: https://issues.redhat.com/browse/OSPRH-15198


abays (approved):
/lgtm

PR #559, opened 2025-04-04
Import ExtraMounts to keystone-operator

fmount:
This patch introduces the `extraMounts` feature and the `Propagation` keys provided by storage.
The purpose is to allow entities to provide files (via `Secrets`, `ConfigMaps`, regular `PVCs` and so forth) to `KeystoneAPI` pods via the existing `Volume/VolumeMounts` interface.


xek (approved):
Thanks!

PR #563, opened 2025-04-11
Enable MTLS memcached auth

lmiccini:
This commit allows users to use a pre-created certificate stored in a secret to perform mtls authentication against the memcached instance(s).

Setting "sslVerifyMode: Request" in the openstackcontrolplane cr :

~~~
  memcached:
    enabled: true
    templates:
      memcached:
         cacheSize: 9932
         maxConn: 8192
         replicas: 1
         tls:
           mtls:
             authCertSecret: {}
             sslVerifyMode: Request 
~~~

results in:

~~~
$ oc get memcached memcached -o yaml 
...
apiVersion: v1
items:
- apiVersion: memcached.openstack.org/v1beta1
  kind: Memcached
  spec:
  ...
    tls:
      caBundleSecretName: combined-ca-bundle
      mtls:
         authCertSecret:
           secretName: cert-memcached-mtls
         sslVerifyMode: Request
      secretName: cert-memcached-svc
status:
  conditions:
~~~

and keystone.conf to be updated like:

~~~
[cache]

backend = dogpile.cache.pymemcache
memcache_servers=memcached-0.memcached.openstack.svc:11212,memcached-1.memcached.openstack.svc:11212,memcached-2.memcached.openstack.svc:11212                            
enable_retry_client = true
retry_attempts = 2
retry_delay = 0

tls_certfile=/etc/pki/tls/certs/mtls.crt
tls_keyfile=/etc/pki/tls/private/mtls.key
tls_cafile=/etc/pki/tls/certs/mtls-ca.crt


enabled=true
tls_enabled=true
~~~

Note: setting sslVerifyMode to "Request" will not remove the non-tls memcache svc (11211), while setting it to "Require" will, so that gives us a staged approach to enabling it without breaking currently running services.

Jira: https://issues.redhat.com/browse/OSPRH-14385


controllers/keystoneapi_controller.go
```diff
@@ -1290,6 +1290,17 @@ func (r *KeystoneAPIReconciler) generateServiceConfigMaps(
...
 
+	// MTLS params
+	if mc.Status.MTLSCert != "" {
+		templateParameters["MemcachedAuthCert"] = fmt.Sprint(memcachedv1.CertMountPath())
+		templateParameters["MemcachedAuthKey"] = fmt.Sprint(memcachedv1.KeyMountPath())
+		templateParameters["MemcachedAuthCa"] = fmt.Sprint(memcachedv1.CaMountPath())
+	} else {
+		templateParameters["MemcachedAuthCert"] = ""
+		templateParameters["MemcachedAuthKey"] = ""
+		templateParameters["MemcachedAuthCa"] = ""
```
xek, line 1301: Could this break the current configuration? If I understand this correctly, it removes the TLS cert, key and ca, if the mTLS cert is not provided, which currently, even with the TLS Everywhere enabled configuration, wouldn't be present.

lmiccini: as far as I know the current TLS-e keystone.conf is like this:

`[cache]

backend = dogpile.cache.pymemcache
memcache_servers=memcached-0.memcached.openstack.svc:11212,memcached-1.memcached.openstack.svc:11212,memcached-2.memcached.openstack.svc:11212
enable_retry_client = true
retry_attempts = 2
retry_delay = 0

enabled=true
tls_enabled=true`

and the CA certificates used are the ones under /etc/pki/tls etc.

This patch is only additive as it adds to the [cache] sections are as per https://docs.openstack.org/keystone/latest/configuration/samples/keystone-conf.html to perform the client auth if required.
I hope I am not missing anything.

stuggi: I think we could remove the else when we do `{{ if (index . "MemcachedAuthCert") }}` in the template as the condition.

fmount (approved):
/lgtm

stuggi (approved):
/lgtm

PR #565, opened 2025-04-14
Optmize handling topology logic

fmount:
This patch removes the need of using `GetLastAppliedTopologyRef` to build a `topologyRef` used by `EnsureServiceTopology` provided by infra-operator. In particular, we can now blindly pass `instance.Status.LastAppliedTopologyRef`, and checking that is not nil is deferred to the infra-operator logic [1]. This change also allows to deduplicate most of the common code present in all operators.
As part of the same change, envTests are improved and both `CreateTopology` and `GetTopology` function calls now rely on infra-operator's implementation. 

Depends-On: https://github.com/openstack-k8s-operators/infra-operator/pull/367
Jira: https://issues.redhat.com/browse/OSPRH-15412

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/367


stuggi (approved):
/lgtm

PR #567, opened 2025-04-16
[WIP] AppCred controller support

Deydra71:
This PR introduces a new ApplicationCredential (AC) controller in the keystone-operator. It watches `ApplicationCredential` custom resources and performs these actions:

1. Creates Keystone ApplicationCredentials for each CR (authenticating as that user due to Keystone’s default policy)
2. Stores the AC’s ID and Secret in a k8s secret
3. Implements rotation logic based on `expirationDays` and `gracePeriodDays`:
   - Reconciles at least once a day, rotating any AC that’s within or past its grace window
   - If an AC is already in the grace period at the next reconcile, it rotates immediately
   - The old ApplicationCredential in Keystone is *not revoked* on rotation (it naturally expires)

Additionally:
- Deletion of an `ApplicationCredential` CR revokes the credential immediately in Keystone
- The controller waits for a `KeystoneAPI` resource to be `Ready` before proceeding with AC operations

Notes:
- CRD & RBAC for the `ApplicationCredential` resource are not automatically installed yet. These must be applied manually until openstack-operator integration is complete

To apply rbac permissions run `oc edit clusterrole keystone-operator-manager-role` and add:
```
- apiGroups:
  - keystone.openstack.org
  resources:
  - applicationcredentials
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch

- apiGroups:
  - keystone.openstack.org
  resources:
  - applicationcredentials/finalizers
  verbs:
  - patch
  - update

- apiGroups:
  - keystone.openstack.org
  resources:
  - applicationcredentials/status
  verbs:
  - get
  - patch
  - update
```
Example AC CR for barbican service user:
```
kind: ApplicationCredential
metadata:
  name: ac-barbican
  namespace: openstack
spec:
  userName: barbican
  expirationDays: 2
  gracePeriodDays: 1
```


api/v1beta1/keystoneapi.go
```diff
@@ -92,6 +93,115 @@ func GetAdminServiceClient(
...
+}
+
+func getPasswordFromOSPSecret(
+	ctx context.Context,
+	h *helper.Helper,
+	userName string,
+) (string, error) {
+
+	// The name of the Secret containing the service passwords
+	const ospSecretName = "osp-secret"
```
vakwetu, line 181: We've discussed this - but basically we provide the ability for the operators to specify the secret that contains the admin user password.  This is osp-secret by default - but it need not be.  See https://github.com/openstack-k8s-operators/barbican-operator/blob/main/api/v1beta1/common_types.go#L44  for instance.

api/v1beta1/keystoneapi.go
```diff
@@ -92,6 +93,115 @@ func GetAdminServiceClient(
...
+
+	secretObj, _, err := secret.GetSecret(ctx, h, ospSecretName, h.GetBeforeObject().GetNamespace())
+	if err != nil {
+		return "", fmt.Errorf("failed to get Secret/%s: %w", ospSecretName, err)
+	}
+
+	key := capitalizeFirst(userName) + "Password"
```
vakwetu, line 188: This may be a default, but its not what is specified.  In barbican, for instance we have a PasswordSelectors field - https://github.com/openstack-k8s-operators/barbican-operator/blob/main/api/v1beta1/common_types.go#L49 which identifies the correct key.  But, this needn't be the case.

Ultimately, I think you are going to need to have the AC specification include the name of the user, the name of the user secret and the relevant field.  You can set these appropriately in openstack-operator.

api/v1beta1/keystoneapplicationcredential_types.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:validation:XValidation:rule="self.gracePeriodDays < self.expirationDays",message="gracePeriodDays must be smaller than expirationDays"
+// ApplicationCredentialSpec defines what the user can set
+type ApplicationCredentialSpec struct {
+	// UserName - the Keystone user under which this AC is created
+	UserName string `json:"userName"`
```
vakwetu, line 28: As mentioned above, you probably need the secretName and field as addtional parameters.

api/v1beta1/keystoneapplicationcredential_types.go
```diff
@@ -0,0 +1,94 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=14
+	// +kubebuilder:validation:Minimum=2
+	ExpirationDays int `json:"expirationDays"`
+
+	// GracePeriodDays sets how many days before expiration the AC should be rotated
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=7
+	// +kubebuilder:validation:Minimum=1
+	GracePeriodDays int `json:"gracePeriodDays"`
```
vakwetu, line 40: Do we want to add a field here for the specification of access rules?  That will be the next logical extension of this work.  Better to get the API change in now in the same release.  See access rules under https://docs.openstack.org/keystone/latest/user/application_credentials.html

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	createOpts := applicationcredentials.CreateOpts{
+		Name:         newACName,
+		Description:  fmt.Sprintf("Created by keystone-operator for user %s", ac.Spec.UserName),
+		Unrestricted: false,
+
+		// Always assign these roles:
+		Roles: []applicationcredentials.Role{
+			{Name: "admin"},
+			{Name: "service"},
+		},
```
vakwetu, line 283: This makes this controller less generic and therefore potentially less useful.   Perhaps this is another parameter - like the access rules that should be passed in as part of the AC spec.

The same can also be said for the Unrestricted field.

vakwetu: I see there is support for all of Roles, Unrestricted and AccessRules in the gophercloud call - https://pkg.go.dev/github.com/gophercloud/gophercloud/openstack/identity/v3/applicationcredentials#CreateOpts

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	grace := time.Duration(ac.Spec.GracePeriodDays) * 24 * time.Hour
+	rotateAt := expiry.Add(-grace)
+
+	if rotateAt.Before(time.Now()) {
+		// Already within the grace window, so trigger rotation immediately
+		return 0
+	}
+
+	// Otherwise check again in 24 hours
+	return defaultRequeue
```
vakwetu, line 379: Shouldn't this be return rotateAt ?

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	rotateAt := expiry.Add(-grace)
+	return time.Now().After(rotateAt)
+}
+
+// computeNextRequeue schedules the next check to happen at "expiry - grace"
+func (r *ApplicationCredentialReconciler) computeNextRequeue(ac *keystonev1.ApplicationCredential) time.Duration {
+	// default requeue is 24h as minimal grace period is 1 day
+	defaultRequeue := 24 * time.Hour
+	if ac.Status.ExpiresAt == nil || ac.Status.ExpiresAt.IsZero() {
+		return defaultRequeue
```
vakwetu, line 365: I'm curious - why would we want to requeue if the application credential does not expire?

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+// randomSuffix generates a random string suffix
+func randomSuffix(n int) string {
+	b := make([]byte, (n+1)/2)
+	_, err := rand.Read(b)
+	if err != nil {
+		return "fallback" // placeholder for generating failure
+	}
+	s := hex.EncodeToString(b)
+	return s[:n]
```
vakwetu, line 390: Its unlikely, but should we check for collisions?

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+func (r *ApplicationCredentialReconciler) reconcileDelete(
+	ctx context.Context,
+	instance *keystonev1.ApplicationCredential,
+	helperObj *helper.Helper,
+) (ctrl.Result, error) {
+
+	logger := r.GetLogger(ctx)
+
+	// Only if user explicitly does "oc delete" do we revoke the AC
```
vakwetu, line 237: Its actually quite useful, I think, to have the delete revoke the application credential.  This gives us a nice way to do a revocation.

The problem is, of course, that there will be some time between when the app cred is revoked and the new one is issued.

I wonder if there is  a way to trigger a rotation without doing a delete.  Could we implement a reconcileUpdate to do this instead?  Then, the procedure when trying to revoke the cert would be to -
1. patch the AC so that we are within the grace period.  This triggers the creation of a new AC.
2. Wait for the new AC to be propagated.
3. Revoke the old AC by deleting it.

controllers/keystoneapplicationcredential_controller.go
Deydra71: `rotateAt ` is a timestamp, we need return a duration

controllers/keystoneapplicationcredential_controller.go
Deydra71: I was thinking about this as an ultimate fallback to wake the controller at least once a day. If because of some error the status of the CR is not updated.

controllers/keystoneapplicationcredential_controller.go
Deydra71: I'm just concerned about step 2. - Wait for the new AC to be propagated. Because AC controller would need a feedback that the service deployment successfully rolled out with new credential, which would add another logic to watch the deployment status in AC controller. I’d prefer to leave revocation as a manual step for now.

api/v1beta1/keystoneapplicationcredential_types.go
mauricioharley: Because of security reasons, I'd also recommend establishing a maximum value.

stuggi: with expire of 14 days and grace 7 days, this will result in rolling restart of all services every week. seems too much for me as a default. shouldn't be the default to rotate like once a year or once every half year max?

What security reason is a max value good for?

api/v1beta1/keystoneapi.go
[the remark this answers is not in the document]
Deydra71: This was just to filter user password from osp-secret. Now, passwordSelector will be passed to AC CR, so that will be used for filtering

api/v1beta1/keystoneapplicationcredential_types.go
Deydra71: Yes, defaults will be changed. I Don't think we need max value, I'm working on a way to trigger rotation manually if necessary.

Deydra71:
The latest update makes the controller take into consideration the custom password secret, custom service user name, and password selectors. Continuing to address other reviews.

api/v1beta1/keystoneapplicationcredential_types.go
mauricioharley: Well, this is a security credential.  The maximum value is to avoid relapse cases when administrators simply forget or are not worried about protecting their credentials.

Not specifying a maximum value in days, i.e. not having a logic to automatically rotate the credential, would leave this responsibility solelly with the sysop/admin.

api/v1beta1/keystoneapi.go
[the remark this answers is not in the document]
mauricioharley: I don't get it.  Can you elaborate?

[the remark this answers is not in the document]
Deydra71: The previous code assumed this password scheme in osp-secret:
```
BarbicanPassword: <secret_data>
CinderPassword: <secret_data>
GlancePassword: <secret_data>
…
```
So to pick the right field we had to:
1. Take the userName ("barbican")
2. Capitalize the first letter → "Barbican"
3. Append "Password" → "BarbicanPassword"

and then look up that key in the secret. Of course that'd force everyone to use that convention. So now openstack-operator extracts and passes into AC CR these as well:
```
spec:
  secret: # the Secret name (by default it's osp-secret)
  passwordSelector:  # how we extract this key, e.g. BarbicanPassword
  userName: # e.g. barbican, user can customize this in control plane spec
  …
```

api/v1beta1/keystoneapplicationcredential_types.go
Deydra71: Not having max value set doesn't equal not rotating. The default expiration and grace periods are always applied, which means we will never end with never expiring ACs.  I change the defaults to what Ade and Martin suggested - 1 year for expiration and 6 months for grace period. 

Putting max here would lock every user to use what we think is optimal, which can vary depending of the purpose of the system.

Deydra71:
Latest update includes corrections based on some reviews (not all, will still continue), and also adds support for the `Unrestricted`, `Roles` and `AccessRoles` fields. 

Currently `AccessRoles` field is not correctly passed to the AC CR by openstack-operator, so it's missing in the AC CR, and for services this field is specified in the openstackcontrolplane spec,  ACs are not generated. We are continuing on Slack what should be in [AccessRules](https://github.com/gophercloud/gophercloud/blob/v1.14.1/openstack/identity/v3/applicationcredentials/results.go#L21) fields in the first place.

api/v1beta1/keystoneapi.go
```diff
@@ -92,6 +92,107 @@ func GetAdminServiceClient(
...
+
+	password, err := getPasswordFromOSPSecret(ctx, h, secretName, passwordSelector)
+	if err != nil {
+		return nil, ctrl.Result{}, fmt.Errorf("failed to get password from osp-secret for user %q: %w", userName, err)
+	}
+
+	scope := &gophercloud.AuthScope{
+		ProjectName: "service",
+		DomainName:  "Default",
+	}
```
vakwetu, line 145: I'm very wary about hard-coding stuff here.  I think that; 
1) there may be cases where either the service project or DomainName may be different
2) we might be interested in obtaining an app cred for a non-service user or for a different project.  For example, I know that there is work ongoing upstream to do manila share encryption where the user would use app creds to retrieve a barbican secret.  

In any case, I think it makes sense to make this function more generic - maybe call it GetUserClient() and pass in the domain and project name.  You could then add these parameters to the app cred spec, and default to "service" and "Default".  This will future-proof the interface a bit.

api/v1beta1/keystoneapi.go
```diff
@@ -92,6 +92,107 @@ func GetAdminServiceClient(
...
+
+	osClient, err := openstack.NewOpenStack(
+		h.GetLogger(),
+		openstack.AuthOpts{
+			AuthURL:    authURL,
+			Username:   userName,
+			Password:   password,
+			TenantName: "service",
```
vakwetu, line 153: ditto comment as above.

controllers/keystoneapplicationcredential_controller.go
vakwetu: OK

controllers/keystoneapplicationcredential_controller.go
vakwetu: OK that makes sense.  I (mis)understood this function when I first read it.

So if I understand this correctly now, we should be returning a recheck duration of 24 hours, unless we are already in the grace period - in which case we would be returning a 0 to immediately recheck.

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,435 @@
...
+			))
+			return ctrl.Result{}, err
+		}
+		logger.Info("Created new AC in Keystone", "ACID", newID, "expiresAt", expiresAt)
+
+		// Store it in a Secret
+		secretName := fmt.Sprintf("%s-secret", instance.Name)
+		if err := r.storeACSecret(ctx, helperObj, instance, secretName, newID, newSecret); err != nil {
+			return ctrl.Result{}, err
+		}
```
vakwetu, line 216: The keystone-operator is somewhat unique in the operators in that there isn't a lot of documentation of what it does.  Just looking at the code , for instance, it wasn't completely clear to me what the name of the ac would be, and the name of the secret etc.  I'm sure I could figure it out - but I shouldn't have to work so hard.

It would be great if there were a small doc (maybe in a docs directory like the other operators) which contains a description of what we'd expect this operator to produce.  It doesn't have to be fancy, but it will help future reviewers and coders who want to enhance this feature.

api/v1beta1/keystoneapi.go
Deydra71: I understand, but right now we don't have any `spec.ServiceProject` or `spec.ServiceDomain`. So, that's why this serves as a mean to only get scoped token for our service.

Since FR3 is explicitly about wiring up service-account auth, I suggest we leave this helper as is for now, and add a // TODO  pointing at the future work to extend the CRD with project/domain fields and refactor into a generic GetUserClient(…, project, domain…)

The domain name is actually hard coded for admin as well - [keystoneapi.go](https://github.com/openstack-k8s-operators/keystone-operator/blob/main/api/v1beta1/keystoneapi.go#:~:text=%2C-,DomainName,-%3A)

Deydra71: > there may be cases where either the service project or DomainName may be different
Is this still true if we take into account only service users?

Deydra71:
The user ID needed to create application credential is now retrieved from the user auth token, so the admin user is no longer required to retrieve user ID.

api/v1beta1/keystoneapplicationcredential_types.go
```diff
@@ -0,0 +1,132 @@
...
+	// +kubebuilder:validation:MinItems=1
+	Roles []string `json:"roles,omitempty"`
+
+	// Unrestricted indicates whether the ApplicationCredential may be used to create or destroy other credentials or trusts
+	// +kubebuilder:validation:Optional
+	Unrestricted bool `json:"unrestricted,omitempty"`
+
+	// AccessRules defines which services the ApplicationCredential is permitted to access
+	// +kubebuilder:validation:Optional
+	AccessRules []ACRule `json:"accessRules,omitempty"`
```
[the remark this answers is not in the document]
Deydra71: They are optional, if not provided the AC will have access based on provided role

api/v1beta1/keystoneapplicationcredential_types.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// GracePeriodDays sets how many days before expiration the ApplicationCredential should be rotated
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=182
+	// +kubebuilder:validation:Minimum=1
+	GracePeriodDays int `json:"gracePeriodDays"`
+
+	// Roles to assign to the ApplicationCredential
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinItems=1
```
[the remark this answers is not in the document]
stuggi: yes, if the parameters are required, I think we should set them as required.

controllers/keystoneapplicationcredential_controller.go
fmount: should this function be a lib-common utility in the long term?

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,517 @@
...
+	if rotateAt.Before(time.Now()) {
+		// already within the grace window, so trigger rotation immediately
+		return 0
+	}
+
+	// otherwise check again in 24 hours
+	return defaultRequeue
+}
+
+// needsRotation returns (shouldRotate, logMessage)
```
fmount, line 479: maybe here we can explain better the returned parameters.
In addition I'm not sure you need both `bool` and `string`.
The string seems something that feeds the logger, while the boolean is what you need to process for real as a returned value, right? In any case, I'm not asking to change this function, I was just trying to see it in the picture of the whole flow.

controllers/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,517 @@
...
+			}
+			return err
+		}
+	}
+
+	return nil
+}
+
+// getUserIDFromToken extracts the user ID from the authenticated token
+func (r *ApplicationCredentialReconciler) getUserIDFromToken(_ logr.Logger, identClient *gophercloud.ServiceClient, username string) (string, error) {
```
fmount, line 442: do you need logger here?

docs/applicationcredentials.md
```diff
@@ -0,0 +1,212 @@
...
+
+// Get standard AC Secret name for a service
+secretName := keystonev1.GetACSecretName("barbican") // Returns "ac-barbican-secret"
+
+// Get standard AC CR name for a service
+crName := keystonev1.GetACCRName("barbican") // Returns "ac-barbican"
+
+// Fetch AC data directly from the Secret
+acData, err := keystonev1.GetApplicationCredentialFromSecret(
+    ctx, client, namespace, "barbican")
```
fmount, line 197: should this be, instead of "barbican", `secretName` that we got on L190 (and therefore `ac-barbican-secret`)?

Deydra71:
> why does this have do-not-merge label?

Thanks, it was a remnant when stuff fro FR4 were getting merged

api/v1beta1/keystoneapplicationcredential_types.go
```diff
@@ -0,0 +1,142 @@
...
+	// +kubebuilder:validation:Optional
+	Service string `json:"service,omitempty"`
+
+	// Path is the API path to allow
+	// +kubebuilder:validation:Optional
+	Path string `json:"path,omitempty"`
+
+	// Method is the HTTP verb to allow (defaults to all if empty)
+	// +kubebuilder:validation:Optional
+	Method string `json:"method,omitempty"`
```
stuggi: if `AccessRules` are specified, is any of these required, or all optional?

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,543 @@
...
+			if k8s_errors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	// Initialize conditions on first pass
+	if instance.Status.Conditions == nil {
```
stuggi: shouldn't we reset the conditions to init on each reconcile, like we do in other operators?

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,543 @@
...
+	nextCheck := r.computeNextRequeue(instance)
+	return ctrl.Result{RequeueAfter: nextCheck}, nil
+}
+
+func (r *ApplicationCredentialReconciler) reconcileDelete(
+	ctx context.Context,
+	instance *keystonev1.KeystoneApplicationCredential,
+) (ctrl.Result, error) {
+
+	// Before removing our CR finalizer, allow ApplicationCredential Secret to be deleted by removing its protection finalizer
```
stuggi: don't we first have to clean it up in keystone by doing an keystoneapi call, similar to how we do it in the keustoneendpoint https://github.com/openstack-k8s-operators/keystone-operator/blob/main/internal/controller/keystoneendpoint_controller.go#L255 ? otherwise we end up in the AC still be there in keystone? or is there some automatic cleanup in keystone and we just let it expire?

api/v1beta1/keystoneapplicationcredential.go
```diff
@@ -0,0 +1,126 @@
...
+		ctx,
+		secretKey,
+		[]string{"AC_ID", "AC_SECRET"},
+		c,
+		timeout,
+	)
+
+	// VerifySecret returns res.RequeueAfter > 0 when secret not found (not an error)
+	// For AC, this is optional, so we just skip it instead of requeueing
+	if res.RequeueAfter > 0 {
```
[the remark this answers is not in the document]
Deydra71: That's a really good catch I didn't notice... With the current sctructure passing timeout equal to 0 breaks the logic if the secret is not found, which means that `verifySecret` returns `res == ctrl.Result{RequeueAfter: timeout}` we skip the `res.RequeueAfter > 0 ` check, so we end up with empty hash. We can change it to:

```
if hash == "" {
        // AC secret not found - this is OK, service will use password auth
        return ctrl.Result{}, nil
    }
```

And not rely on the requeue value returned. Does it make sense?

api/v1beta1/keystoneapi.go
```diff
@@ -138,6 +138,112 @@ func GetAdminServiceClient(
...
+func getPasswordFromOSPSecret(
+	ctx context.Context,
+	h *helper.Helper,
+	ospSecretName, passwordSelector string,
+) (string, ctrl.Result, error) {
+	if ospSecretName == "" {
+		return "", ctrl.Result{}, fmt.Errorf("secret name is empty, cannot retrieve password for selector %q", passwordSelector)
+	}
+	if passwordSelector == "" {
+		return "", ctrl.Result{}, fmt.Errorf("password selector is empty, cannot retrieve password from secret %q", ospSecretName)
```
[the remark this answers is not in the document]
Deydra71: Yes, agree, will add the kube validations for `PasswordSelector` and `Secret` fields.

api/v1beta1/keystoneapplicationcredential.go
```diff
@@ -0,0 +1,133 @@
...
+func GetApplicationCredentialFromSecret(
+	ctx context.Context,
+	c client.Client,
+	namespace string,
+	serviceName string,
+) (*ApplicationCredentialData, error) {
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{Namespace: namespace, Name: GetACSecretName(serviceName)}
+	if err := c.Get(ctx, key, secret); err != nil {
+		if k8s_errors.IsNotFound(err) {
```
fmount: @Deydra71 just wondering if `VerifyApplicationCredentialsForService` should be embedded here since we pass pretty much the same parameters and it would hide more details from the service operators that right now call both functions.
It would be nice to have a single entrypoint, and delegate keystone to do the verification, that of course can return an error we can handle.

docs/applicationcredentials.md
```diff
@@ -0,0 +1,224 @@
...
+  - Sets `.status.lastRotated` to current timestamp
+  - Re-marks AC CR ready
+  - Emits an event to notify EDPM nodes about the rotation
+- Requeue
+  - Schedules the next check at `(newExpiresAt - gracePeriodDays)`
+
+## Manual Rotation
+Manual rotation can be triggered by patching the AC CR with an expiration timestamp in the past, e.g.:
+
+`oc patch -n openstack keystoneapplicationcredential ac-barbican --type=merge --subresource=status -p '{"status":{"expiresAt":"2001-05-19T00:00:00Z"}}'`
```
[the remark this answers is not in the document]
Deydra71: When we delete the expireAt field it doesn't trigger rotation, because `needsRotation()` https://github.com/Deydra71/keystone-operator/blob/0695a0cf6ac9d70f1160571616dc367040f90ee2/internal/controller/keystoneapplicationcredential_controller.go#L524 returns false.

Actually deleting this param may prevent from future rotation as well, because status fields are not reconciled, right? So it will be still missing in the CR until manually is patched again. In this case, should we restore to the latest expiresAt by fetching the data from keystone to we prevent no rotation scenario?

Deydra71:
@stuggi @fmount hi! Could you please take a look again at the updated PR with the suggested changes? We’d like to merge it so we can proceed with kuttl and other testing in the service operators. I also updated the service operators that are supposed to use AppCred (telemetry still in progres) to use field indexers instead of custom watchers.

api/v1beta1/keystoneapi.go
```diff
@@ -138,6 +138,106 @@ func GetAdminServiceClient(
...
+) (string, ctrl.Result, error) {
+	data, res, err := secret.GetDataFromSecret(
+		ctx,
+		h,
+		ospSecretName,
+		10*time.Second,
+		passwordSelector,
+	)
+	if err != nil {
+		return "", ctrl.Result{}, fmt.Errorf("failed to get %q from Secret/%s: %w", passwordSelector, ospSecretName, err)
```
stuggi: do we need this message, the lib-common func already returns an error string depending on the condition it checks ( generic error, secret not there, key not found)

api/v1beta1/keystoneapi.go
```diff
@@ -138,6 +138,106 @@ func GetAdminServiceClient(
...
+	}
+	if res != (ctrl.Result{}) {
+		return "", res, nil
+	}
+	if data == "" {
```
stuggi: do we need this? is an empty string a valid pwd in keystone? creating the authclient should fail.

docs/applicationcredentials.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Rotation
+When the next reconcile hits the grace window (`now ≥ expiresAt - gracePeriodDays`), `needsRotation()` returns `true` again and the controller:
+
+- Service-Scoped Client - authenticates as the service user
+- Extract User ID from the service user's token
+- Create New AC
+  - Generates a new Keystone AC with a fresh 5-char suffix
+  - Uses the same roles, unrestricted flag, access rules, and expirationDays
+  - Does _not_ revoke the old AC, the old credential naturally expires
```
stuggi: what happens when it expired. gets it removed from keystone db automatic, or do we need a cron which cleans them, like we have for the tokens?

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,547 @@
...
+	savedConditions := instance.Status.Conditions.DeepCopy()
+
+	// Always patch the instance status when exiting this function so we can persist any changes.
+	defer func() {
+		if instance.DeletionTimestamp.IsZero() {
+			// Update the Ready condition based on the sub conditions
+			// ApplicationCredentialReady being True implies KeystoneAPIReady is True,
+			// so when ApplicationCredential is ready, setup is complete
+			if instance.Status.Conditions.IsTrue(keystonev1.KeystoneApplicationCredentialReadyCondition) {
+				instance.Status.Conditions.MarkTrue(condition.ReadyCondition, keystonev1.KeystoneApplicationCredentialReadyMessage)
```
stuggi: should be the generic condition.ReadyMessage.

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,547 @@
...
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			keystonev1.KeystoneAPIReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			"KeystoneAPI not found: %v",
+			err,
+		))
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+	}
```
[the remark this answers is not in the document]
stuggi: after that should also call the cleanup as in https://github.com/openstack-k8s-operators/keystone-operator/blob/main/internal/controller/keystoneendpoint_controller.go#L182-L198 for the cases where we do not need the keystoneapi

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,547 @@
...
+	if !instance.DeletionTimestamp.IsZero() && !keystoneAPI.DeletionTimestamp.IsZero() {
+		if controllerutil.ContainsFinalizer(instance, finalizer) {
+			base := instance.DeepCopy()
+			controllerutil.RemoveFinalizer(instance, finalizer)
+			if err := r.Patch(ctx, instance, client.MergeFrom(base)); err != nil {
+				if k8s_errors.IsConflict(err) {
+					return ctrl.Result{Requeue: true}, nil
+				}
+				return ctrl.Result{}, err
+			}
```
stuggi: just return after we removed the finalizer is enough since the defer will do the patch. I think we don't need this condition here and just call reconcileDelete and pass in the keystonAPI for any condition required?

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,547 @@
...
+				userErr.Error(),
+			))
+			return userRes, userErr
+		}
+		if userRes != (ctrl.Result{}) {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				keystonev1.KeystoneApplicationCredentialReadyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				"Waiting for secret %q to be available",
```
stuggi: lets add a KeystoneApplicationCredentialWaitingMessage

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,547 @@
...
+		if isRotation {
+			now := metav1.Now()
+			instance.Status.LastRotated = &now
+
+			// Emit event for EDPM visibility
+			r.EventRecorder.Event(
+				instance,
+				corev1.EventTypeNormal,
+				"ApplicationCredentialRotated",
+				fmt.Sprintf("ApplicationCredential '%s' (user: %s) rotated - EDPM nodes may need credential updates. Previous expiration: %s, New expiration: %s",
```
stuggi: nice! would be also good to set a warning condition https://github.com/openstack-k8s-operators/dev-docs/blob/main/conditions.md#severity on the ctlpane, that the controlplane flips to not ready?

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,547 @@
...
+
+	// Remove finalizer from the ApplicationCredential CR
+	base := instance.DeepCopy()
+	controllerutil.RemoveFinalizer(instance, finalizer)
+	if err := r.Patch(ctx, instance, client.MergeFrom(base)); err != nil {
+		if k8s_errors.IsConflict(err) {
+			return ctrl.Result{Requeue: true}, nil
+		}
+		return ctrl.Result{}, err
+	}
```
stuggi: since it is on the instance itself and return afterwards, the patch is not required since we will do it in the defer func

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -0,0 +1,547 @@
...
+func (r *ApplicationCredentialReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&keystonev1.KeystoneApplicationCredential{}).
+		Owns(&corev1.Secret{}).
+		Complete(r)
+}
+
+// GetLogger returns a logger configured for this controller.
+func (r *ApplicationCredentialReconciler) GetLogger(ctx context.Context) logr.Logger {
+	return ctrlLog.FromContext(ctx).WithName("ApplicationCredentialReconciler")
```
stuggi: we usually use the format:
```
	return ctrlLog.FromContext(ctx).WithName("Controllers").WithName("KeystoneApplicationCredential")
```

internal/controller/keystoneapplicationcredential_controller.go
stuggi: yes that was my idea this morning, but thinking of it again, we probably can not, since we do not know when the edpm side got updated to flip it back to ready. 

@slagle is there a way to check if a nodedataset (all nodes) got updated with the latest config?

api/v1beta1/keystoneapi.go
fmount: I agree about having a generic function, the old one can be deprecated and it might produce a warning, so we would know in the service operators that we need to update to the new function call (e.g. [1]).

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/internal/controller/glance_controller.go#L1126

Deydra71:
ATM not sure why the functional test fails, locally the suite passes, and also the failure is connected to the MariaDB part, not our app cred changes https://github.com/openstack-k8s-operators/keystone-operator/blob/main/test/functional/keystoneapi_controller_test.go#L2246

Deydra71:
@stuggi @fmount I refactored generic GetScopedClient + deprecate GetScopedAdminServiceClient based on your comments, can you please recheck it?

api/v1beta1/keystoneapi.go
```diff
@@ -123,22 +123,164 @@ func GetAdminServiceClient(
...
+	region string,
+	scope *gophercloud.AuthScope,
+) (*openstack.OpenStack, ctrl.Result, error) {
+
+	// Use keystoneAPI region as default if not specified
+	if region == "" {
+		region = keystoneAPI.Spec.Region
+	}
+
+	authURL, err := keystoneAPI.GetEndpoint(endpoint.EndpointInternal)
```
stuggi: does the caller need to be able to specify which endpoint should be used? iirc there might be the need for this for skmo to use the public endpoint?

stuggi (commented):
I think this is now good to go. since I am not fully recovered from being sick, @fmount @abays could I ask you to also have final look please :)

api/v1beta1/keystoneapplicationcredential_types.go
```diff
@@ -0,0 +1,154 @@
...
+*/
+
+package v1beta1
+
+import (
+	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:validation:XValidation:rule="self.gracePeriodDays < self.expirationDays",message="gracePeriodDays must be smaller than expirationDays"
```
fmount: @Deydra71 I have a quick question on L24. This is useful to validate the CR at creation time, and the server would block the user based on the rule defined here.
What happen in case of updates? 
Would it make sense to have a function like:
```golang
 func (r *ApplicationCredentialReconciler) validateSpec(ctx context.Context, ac *keystonev1.ApplicationCredential) error {
      if ac.Spec.GracePeriodDays >= ac.Spec.ExpirationDays {
          // Set error condition, log, and requeue
          return fmt.Errorf("gracePeriodDays (%d) must be less than expirationDays (%d)",
              ac.Spec.GracePeriodDays, ac.Spec.ExpirationDays)
      }
      return nil
  }
```

(which would be probably a better fit for webhooks, but I didn't check that part). 
Maybe I'm overlooking something in the code, I understand at this point even reviewing is not an easy task considering all the pieces involved here.

Deydra71: @fmount thanks for reviewing! 

By 'updates' you mean if user should apply `oc patch` or `oc apply` ?

I tested patching the existing cinder AC CR:
```
oc -n openstack-kuttl-tests patch keystoneapplicationcredential ac-cinder \
  --type=merge \
  -p '{"spec":{"expirationDays":10,"gracePeriodDays":10}}'
The KeystoneApplicationCredential "ac-cinder" is invalid: spec: Invalid value: "object": gracePeriodDays must be smaller than expirationDays
```

And it was rejected by the API. So, that is correct. Is there another related problem I didn't understand from the comment?

PR #579, opened 2025-05-14
Use GetDefaultMemcachedSpec test helper from infra-operator

ratailor:
This change uses GetDefaultMemcachedSpec test helper from infra-operator to reduce code duplication.


stuggi (approved):
/lgtm

PR #591, opened 2025-06-02
Add KeystoneAPIStatusChangedPredicate to help watch KeystoneAPI status to reconcile on endpoint changes

stuggi:
Adds KeystoneAPIStatusChangedPredicate where its primary purpose is to return true if the KeystoneAPI Status.APIEndpoints has changed.

In addition also returns true if it gets deleted, which is unlikely to happen since keystone is a requirement for other services to work properly. Used by service operators to watch KeystoneAPI endpoint they depend on.
    
Jira: [OSPRH-16994](https://issues.redhat.com//browse/OSPRH-16994)

### DIFF
--- a/api/v1beta1/keystoneapi.go
+++ b/api/v1beta1/keystoneapi.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"reflect"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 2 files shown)


api/v1beta1/keystoneapi.go
```diff
@@ -27,13 +28,41 @@ import (
...
+var KeystoneAPIStatusChangedPredicate = predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		if e.ObjectOld == nil || e.ObjectNew == nil {
+			return false
+		}
+		oldPod, okOld := e.ObjectOld.(*KeystoneAPI)
+		newPod, okNew := e.ObjectNew.(*KeystoneAPI)
+
+		if !okOld || !okNew {
+			return false // Not a keystonev1.KeystoneEndpoint, or cast error
```
dprince, line 53: @stuggi just noticed the comment here references Endpoints. It should say KeystoneAPI right?

stuggi: If you check bellow the deepequal, it is checking the keystoneapi endpoints from the status

PR #597, opened 2025-07-01
Use kolla for mounting multi-realm federation config files

Deydra71:
The original multirealm‐federation patch #592  mounted files straight into /etc/ as root, making them unreadable by the apache and keystone users. This change instead mounts into /var/lib/... first, then copies the files into /etc/httpd/conf/ with the proper ownership.

### DIFF
--- a/templates/keystoneapi/config/keystone-api-config.json
+++ b/templates/keystoneapi/config/keystone-api-config.json
@@ -65,7 +65,15 @@
             "owner": "keystone:apache",
             "perm": "0444",
             "optional": true
+        },
+        {
+            "source": "/var/lib/config-data/default/multirealm-federation/*",
+            "dest": "/etc/httpd/conf/",
+            "owner": "keystone:apache",
+            "perm": "0644",
+            "optional": true
         }
+
     ],
     "permissions": [
         {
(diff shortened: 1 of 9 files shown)


templates/keystoneapi/config/keystone-api-config.json
```diff
@@ -81,7 +81,16 @@
             "owner": "keystone:apache",
             "perm": "0444",
             "optional": true
+        },
+        {
+            "source": "/var/lib/config-data/default/multirealm-federation/*",
+            "dest": "/var/lib/httpd/metadata/",
+            "owner": "keystone:apache",
+            "perm": "0644",
```
vakwetu, line 89: should this be 640?

The whole point behind this change was to have the files owned by keystone:apache - and readable/usable only by those users.  Given that this file contains some passwords (client_password), it probably shouldn't be readable by other.

Incidentally, I do think that its likely that without this patch, the files would have been readable because they were owned by root but had global read permissions.  I  think the problem was that we were trying to test with the CLI - and that just doesn't work.

vakwetu (approved):
LGTM

vakwetu:
Agreed that we should override the pre-commit check.  Not sure how to do that.

Deydra71:
@stuggi  Hello! Can you please assist us with overwriting the pre-commit check? The warning is about spec type that was not used anywhere (the https://github.com/openstack-k8s-operators/keystone-operator/pull/592  introduced it), so it's safe to delete.

PR #603, opened 2025-07-18
Add Region parameter in the keystoneAPI status

fmount:
There are services like `Glance` that need to configure parameters that are based on the `Keystone CR` we build. 
`oslo_limits` is an example of that, and we need to access the `Region` parameter that is defined in `keystone CR` spec.
It might be possible to rely on `gophercloud` and perform `os.GetRegion()`, but it seems too convoluted given that we regularly consume a `keystoneAPI` CR via `GetKeystoneAPI` from the service controllers and we this parameter is built based on the keystoneAPI Spec.
This patch mirrors the `Region` parameter in the `.Status` (according to the `ObservedGeneration` pattern), so we can access this field from a Service operator that needs it through the appropriate API function.

### DIFF
--- a/api/bases/keystone.openstack.org_keystoneapis.yaml
+++ b/api/bases/keystone.openstack.org_keystoneapis.yaml
@@ -1638,6 +1638,9 @@ spec:
                 description: ReadyCount of keystone API instances
                 format: int32
                 type: integer
+              region:
+                description: Region - optional region name for the keystone service
+                type: string
               transportURLSecret:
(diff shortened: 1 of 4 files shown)


fmount:
we have it here [1]

[1] https://logserver.rdoproject.org/b06/rdoproject.org/b0631cd2d992490890b78b091df94271/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/keystoneapis.keystone.openstack.org/keystone.yaml

PR #607, opened 2025-07-28
Add federatedRealmConfig field index

Deydra71:
`federatedRealmConfig` secret is now indexed, so if the secret's string data is changed, operator reconciles.

Using now `GetSecret()` instead of `GetDataFromSecret()`, so the missing `federatedRealmConfig` secret error is shown immediately now, and we just don't silently requeue.

### DIFF
--- a/controllers/keystoneapi_controller.go
+++ b/controllers/keystoneapi_controller.go
@@ -254,6 +254,7 @@ const (
 	tlsAPIPublicField                   = ".spec.tls.api.public.secretName"
 	topologyField                       = ".spec.topologyRef.Name"
 	httpdCustomServiceConfigSecretField = ".spec.httpdCustomization.customServiceConfigSecret" // #nosec G101
+	federatedRealmConfigField           = ".spec.federatedRealmConfig"                         // #nosec G101
 )
 
 var allWatchFields = []string{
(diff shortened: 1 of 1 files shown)


controllers/keystoneapi_controller.go
```diff
@@ -1651,21 +1665,24 @@ func (r *KeystoneAPIReconciler) ensureFederationRealmConfig(
...
+	// Get the data from the federation secret, we expect this to be a JSON dict
+	jsonData, ok := federationSecret.Data[keystone.FederationConfigKey]
+	if !ok {
+		return nil, fmt.Errorf("key %s not found in secret %s: %w", keystone.FederationConfigKey, instance.Spec.FederatedRealmConfig, util.ErrFieldNotFound)
+	}
+
 	// Parse the JSON content into a map
 	var rawConfigs map[string]json.RawMessage
-	err = json.Unmarshal([]byte(jsonData), &rawConfigs)
+	err = json.Unmarshal(jsonData, &rawConfigs)
```
Deydra71, line 1685: Just a note that in  this solution we're pulling directly out of `federationsecret.Data`, which gives  `[]byte` already, so no need to convert.

vakwetu (approved):
LGTM

PR #610, opened 2025-08-11
[OSPRH-18976] Update deprecated Kustomize fields in config dir

abays:
`patchesStrategicMerge` and `patchesJson6902` are deprecated in Kustomize and result in slight-but-impactful differences in the generated output compared to `patches` in certain situations.  We should be using `patches` across all operators instead.

Co-authored-by: Claude (AI Assistant) claude@anthropic.com

Jira: https://issues.redhat.com/browse/OSPRH-18976


stuggi:
lgtm

PR #611, opened 2025-08-13
Add GetKeystoneAPIByName method

ratailor:
This change adds GetKeystoneAPIByName method that takes name of KeystoneAPI as parameter instead of labels and returns KeystoneAPI instance.


ratailor:
this can be used here https://github.com/openstack-k8s-operators/nova-operator/blob/e890fbdf87267967a9b1c47f4446e3fd0edfe593/controllers/nova_controller.go#L1652-L1655 to remove the TODO.

PR #615, opened 2025-08-19
Use quorum queues if enabled

lmiccini:
With https://github.com/openstack-k8s-operators/infra-operator/pull/429 we added the ability to configure RabbitMQ to use quorum queues.

This commit adds a check to verify if "quorumqueues" has been added to the TransportURL secret. If its value is true, the services will be configured to use quorum queues.

Jira: https://issues.redhat.com/browse/OSPRH-19160


stuggi (approved):
/lgtm

PR #621, opened 2025-09-04
Fix Memcached MTLS certs mountpaths for jobs not using kolla

lmiccini:
Keystone cronjob and bootstrap job do not use kolla to copy configuration files and other necessary bits to the proper places, so the mtls ca/key/certs are left under /var/lib and the boostrap or token flush commands would fail. With this commit we now mount certificates directly to the proper places.

Jira: https://issues.redhat.com/browse/OSPRH-19648


olliewalsh (approved):
/lgtm

PR #625, opened 2025-09-11
Add GetKeystoneOverrides function

fmount:
This function is a centralized helper that allows service operators to get a `Secret` based on a `labelSelector` containing `Keystone` overrides. This allows to point services from multiple deployments to the same `keystoneAPI`, and override `auth_url`, `region`, `www_authenticate_uri` parameters.
All the other parameters are ignored because not supported as overrides.

### DIFF
--- a/api/v1beta1/keystoneapi.go
+++ b/api/v1beta1/keystoneapi.go
@@ -36,6 +36,13 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"strings"
+)
+
+var (
+	KeystoneOverridesLabelSelector = map[string]string{
+		"keystone-overrides": "true",
+	}
 )
 
 // KeystoneAPIStatusChangedPredicate - primary purpose is to return true if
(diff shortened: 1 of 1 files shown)


api/v1beta1/keystoneapi.go
```diff
@@ -212,3 +219,48 @@ func GetScopedAdminServiceClient(
...
+		return overrides, fmt.Errorf("could not get secrets: %w", err)
+	}
+	// list is empty, return
+	if len(secrets.Items) == 0 {
+		h.GetLogger().Info("No keystone overrides, using default configuration")
+		return overrides, nil
+	}
+	// Invalid case: more than one secret with the same label is present in the
+	// same namespace. This is not allowed and an error must be returned
+	if len(secrets.Items) > 1 {
```
Deydra71, line 250: I wonder  if we should treat this case differently. Right now, we are not returning any error and when more than one secret with the label are present in the env, the glance reconciles to keystone defaults, no matter what was the previous state. So, if we had overrides already mounted, pods revert to defaults. Maybe in this case we should log the info (as we do now), but not rollout new pods to prevent possible disruption or unnecessary config change?

fmount: Thank you, this is an interesting perspective. It makes sense to me and we can return an error, with the right message, to prevent Glance (or any other service) to reconcile, until the secret is fixed.
It shouldn't happen to see two secrets with the same `LabelSelector`, but I agree that we should preserve the service configuration in this case.

PR #685, opened 2026-04-01
Add finalizer management and EDPM-aware cleanup for immutable AC secrets

Deydra71:
Jira: [OSPRH-28176](https://redhat.atlassian.net/browse/OSPRH-28176), [OSPRH-27512](https://redhat.atlassian.net/browse/OSPRH-27512), [OSPRH-27519](https://redhat.atlassian.net/browse/OSPRH-27519)

Application Credential dev-doc: https://github.com/openstack-k8s-operators/dev-docs/blob/main/application_credentials.md

* Delete unused `GetApplicationCredentialFromSecret` function
* Add a service label to the AC  secret for easy discovery, e.g. `application-credential-service: barbican`
* Introduce immutable per-rotation AC secrets with deterministic names
*  Add a new status field `previousSecretName` used for tracking previously used AC secret 
* Add Keystone-side revocation of unused rotated ACs - currently used secret and previously used secret are protected, pre-previous secret is deleted and revoked
* Suppress Owns() create events on the secret watch to prevent a race condition caused by stale informer cach and sometimes causing additional AC secret to be created and deleted immediately during app cred rotation
* Add migration support for old mutable secrets - on the first reconcile after upgrade, the `application-credential-service ` label is added to any existing secret that lacks it, making it visible to the label-based deletion. This ensures old secrets are properly revoked and cleaned up on the next rotation or CR deletion, and prevents orphaned protection finalizers 
* Add EDPM-aware guards on both cleanup and deletion paths using `edpm.AreSecretHashesInSync()` from https://github.com/openstack-k8s-operators/lib-common/tree/main/modules/edpm. Before revoking rotated AC secrets or proceeding with AC CR deletion, the controller checks that all `OpenStackDataPlaneNodeSet` resources have been redeployed with the current config secrets. If any NodeSet has stale hashes, action is deferred until the next EDPM deploy. The controller watches NodeSet resources (via unstructured access to avoid circular dependencies) and re-evaluates when a NodeSet status changes. All AC CRs are checked uniformly - no hardcoded EDPM service mapping. This approach is aligned with the RabbitMQ user deletion design in infra-operator.

Each service operator that consumes an AC secret now places a `openstack.org/<service>-ac-consumer` finalizer on the AC secret it is actively using. This ensures the keystone-operator cannot revoke or clean up secret while a service is still holding a reference to it.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/693

### DIFF
--- a/api/v1beta1/keystoneapplicationcredential.go
+++ b/api/v1beta1/keystoneapplicationcredential.go
@@ -18,25 +18,21 @@ package v1beta1
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/object"
 )
 
 // ApplicationCredentialData contains AC ID/Secret extracted from a Secret
-// Used by service operators to get AC data from the Secret
 type ApplicationCredentialData struct {
 	ID     string
 	Secret string
-}
-
-// GetACSecretName returns the standard AC Secret name for a service
-func GetACSecretName(serviceName string) string {
-	return fmt.Sprintf("ac-%s-secret", serviceName)
 }
 
 // GetACCRName returns the standard AC CR name for a service
@@ -51,37 +47,68 @@ const (
 	ACSecretSecretKey = "AC_SECRET"
 )
 
-var (
-	// ErrACIDMissing indicates AC_ID key missing or empty in the Secret
-	ErrACIDMissing = errors.New("applicationcredential secret missing AC_ID")
-	// ErrACSecretMissing indicates AC_SECRET key missing or empty in the Secret
-	ErrACSecretMissing = errors.New("applicationcredential secret missing AC_SECRET")
-)
-
-// GetApplicationCredentialFromSecret fetches and validates AC data from the Secret
-func GetApplicationCredentialFromSecret(
+// ManageACSecretFinalizer ensures consumerFinalizer is present on the AC secret
+// identified by newSecretName and absent from the one identified by
+// oldSecretName. It is a no-op when both names are equal.
+func ManageACSecretFinalizer(
 	ctx context.Context,
-	c client.Client,
+	h *helper.Helper,
 	namespace string,
-	serviceName string,
-) (*ApplicationCredentialData, error) {
-	secret := &corev1.Secret{}
-	key := types.NamespacedName{Namespace: namespace, Name: GetACSecretName(serviceName)}
-	if err := c.Get(ctx, key, secret); err != nil {
-		if k8s_errors.IsNotFound(err) {
-			return nil, nil
+	newSecretName string,
+	oldSecretName string,
+	consumerFinalizer string,
+) error {
+	if newSecretName == oldSecretName {
+		return nil
+	}
+
+	var newObj, oldObj client.Object
+
+	if newSecretName != "" {
+		secret := &corev1.Secret{}
+		key := types.NamespacedName{Name: newSecretName, Namespace: namespace}
+		if err := h.GetClient().Get(ctx, key, secret); err != nil {
+			return fmt.Errorf("failed to get new AC secret %s: %w", newSecretName, err)
 		}
-		return nil, fmt.Errorf("get applicationcredential secret %s: %w", key, err)
+		newObj = secret
 	}
 
-	acID, okID := secret.Data[ACIDSecretKey]
-	if !okID || len(acID) == 0 {
-		return nil, fmt.Errorf("%w: %s", ErrACIDMissing, key.String())
-	}
-	acSecret, okSecret := secret.Data[ACSecretSecretKey]
-	if !okSecret || len(acSecret) == 0 {
-		return nil, fmt.Errorf("%w: %s", ErrACSecretMissing, key.String())
+	if oldSecretName != "" {
+		secret := &corev1.Secret{}
+		key := types.NamespacedName{Name: oldSecretName, Namespace: namespace}
+		if err := h.GetClient().Get(ctx, key, secret); err != nil {
+			if !k8s_errors.IsNotFound(err) {
+				return fmt.Errorf("failed to get old AC secret %s: %w", oldSecretName, err)
+			}
+		} else {
+			oldObj = secret
+		}
 	}
 
-	return &ApplicationCredentialData{ID: string(acID), Secret: string(acSecret)}, nil
+	return object.ManageConsumerFinalizer(ctx, h, newObj, oldObj, consumerFinalizer)
+}
+
+// RemoveACSecretConsumerFinalizer removes consumerFinalizer from the AC secret
+// identified by secretName. It is a no-op when secretName is empty or the
+// secret no longer exists.
+func RemoveACSecretConsumerFinalizer(
+	ctx context.Context,
+	h *helper.Helper,
+	namespace string,
+	secretName string,
+	consumerFinalizer string,
+) error {
+	if secretName == "" {
+		return nil
+	}
+
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{Name: secretName, Namespace: namespace}
+	if err := h.GetClient().Get(ctx, key, secret); err != nil {
+		if k8s_errors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	return object.RemoveConsumerFinalizer(ctx, h, secret, consumerFinalizer)
 }
--- a/api/v1beta1/keystoneapplicationcredential.go
+++ b/api/v1beta1/keystoneapplicationcredential.go
@@ -18,25 +18,22 @@ package v1beta1
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/object"
 )
 
 // ApplicationCredentialData contains AC ID/Secret extracted from a Secret
-// Used by service operators to get AC data from the Secret
 type ApplicationCredentialData struct {
 	ID     string
 	Secret string
-}
-
-// GetACSecretName returns the standard AC Secret name for a service
-func GetACSecretName(serviceName string) string {
-	return fmt.Sprintf("ac-%s-secret", serviceName)
 }
 
 // GetACCRName returns the standard AC CR name for a service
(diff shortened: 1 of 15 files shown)


mauricioharley (changes requested):
Thanks for the work on this, Veronika. The overall architecture is solid - immutable per-rotation secrets, Keystone-side revocation, and consumer finalizer coordination are the right design choices.

I left five inline comments on areas that could lead to subtle issues in production:

1. **Infinite requeue loop** (line 223) - unbounded retry when a Secret is genuinely deleted
2. **Fragile substring match** (line 564) - `hasConsumerFinalizer` could match unrelated finalizers
3. **Orphaned secrets** (line 606) - crash between finalizer removal and delete leaves permanently uncleaned secrets
4. **Immutable secret retry failure** (line 729) - `CreateOrPatch` fails on retry against an immutable Secret
5. **Implicit naming convention** (line 584) - `TrimPrefix` silently misbehaves if CR name doesn't follow `ac-<service>`

Please address these before merging.

Deydra71:
Note: I used the `/code-review` skill available through https://github.com/fmount/openstack-k8s-agent-tools/  on the latest changes

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -440,41 +703,76 @@ func (r *ApplicationCredentialReconciler) createACWithName(
...
+
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      secretName,
 			Namespace: ac.Namespace,
+			Labels: map[string]string{
+				"application-credentials":        "true",
+				"application-credential-service": serviceName,
+			},
+			Finalizers: []string{acSecretFinalizer},
```
fmount, line 758: @Deydra71 I see we have both a "producer" finalizer (this line), and a "consumer" finalizer, which is set by the service that uses the secret.
Is it accurate to think that this is useful for `rotation` purposes? In other words I imagine that the consumer moves the finalizer from the old secret to the new one. At that point the old secret has no consumer finalizer, but keystone needs to keep it alive until the AC is revoked. Then it can explicitly remove the
producer finalizer and delete the secret.

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -440,41 +703,76 @@ func (r *ApplicationCredentialReconciler) createACWithName(
...
-		r.GetLogger(ctx).Info("Secret operation completed", "secret", secretName, "operation", op)
+	if err := helperObj.GetClient().Create(ctx, secret); err != nil {
+		if k8s_errors.IsAlreadyExists(err) {
+			// A secret with this name already exists - most likely a previous reconcile
+			// created it but crashed before patching the status. Validate that the
+			// existing secret's ACID matches the one we intended to store, a mismatch
+			// would indicate an ACID prefix collision (two different Keystone AC IDs
+			// whose first 5 characters are identical, producing the same secret name).
+			existing := &corev1.Secret{}
+			if getErr := helperObj.GetClient().Get(ctx, types.NamespacedName{Namespace: ac.Namespace, Name: secretName}, existing); getErr != nil {
```
fmount: wondering if we should rely on lib-common `GetSecret` here [1].
I guess we also have an option to get some data directly [2], so you could get `.Data[keystonev1.ACIDSecretKey]` and compare it here. Not sure this might deserve a dedicated helper but reusing existing functions/patterns might help to keep the code readable.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/secret/secret.go#L77
[2] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/secret/secret.go#L438

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -346,44 +389,264 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+	for _, sn := range []string{instance.Status.SecretName, instance.Status.PreviousSecretName} {
+		if sn != "" {
+			if err := r.ensureServiceLabel(ctx, helperObj, sn, instance.Namespace, serviceName); err != nil {
+				logger.Info("Could not ensure service label on secret during delete", "secret", sn, "error", err)
+			}
+		}
+	}
+
+	secretList := &corev1.SecretList{}
+	if err := helperObj.GetClient().List(ctx, secretList,
```
fmount: It would be useful to have a lib-common function to filter secrets by labels, so we can shorten this code and keep here just the relevant logic. It would be also useful for service operators in case we need to list secrets by label.

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -346,44 +389,264 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+			instance.Spec.Secret,
+			instance.Spec.PasswordSelector,
+		)
+		if userErr != nil || userRes != (ctrl.Result{}) {
+			logger.Info("Could not build Keystone client, skipping revocation during AC CR delete", "error", userErr)
+		} else if userID, err := r.getUserIDFromToken(ctx, userOS.GetOSClient(), instance.Spec.UserName); err != nil {
+			logger.Info("Could not get user ID, skipping revocation during AC CR delete", "error", err)
+		} else {
+			seen := make(map[string]bool)
+			for i := range secretList.Items {
```
fmount: I see we list secrets many times in this function with the purpose of doing some processing and identify which one should be deleted. I'm wondering if we can make this function more linear and merge some processing steps (e.g. we build a list of secrets we can revoke and we run a massive revocation in a single call, maybe with an helper that wraps `revokeKeystoneAC`).
I find hard to follow the logic of this function, and you should consider splitting the logic into smaller pieces that flatten the logic.

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -346,44 +389,264 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+	protected := make(map[string]bool, 2)
+	if instance.Status.SecretName != "" {
+		protected[instance.Status.SecretName] = true
+	}
+	if instance.Status.PreviousSecretName != "" {
+		protected[instance.Status.PreviousSecretName] = true
+	}
+
+	secretList := &corev1.SecretList{}
+	if err := helperObj.GetClient().List(ctx, secretList,
```
fmount: Here we do the same iteration (by label) that can be realized through a lib-common helper (see my previous comment on `reconcileDelete`).

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -346,44 +389,264 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+		acID := string(s.Data[keystonev1.ACIDSecretKey])
+		if acID != "" {
+			if err := revokeKeystoneAC(ctx, identClient, userID, acID); err != nil {
+				return err
+			}
+			logger.Info("Revoked AC in Keystone", "ACID", acID, "secret", s.Name)
+		}
+
+		fresh := &corev1.Secret{}
+		if err := helperObj.GetClient().Get(ctx, types.NamespacedName{Namespace: s.Namespace, Name: s.Name}, fresh); err != nil {
```
fmount: why do we `.Get(` a secret (stored in `s`) that we already have? I'm not sure this logic is required here (L630 - L635)

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -346,44 +389,264 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+			}
+			return err
+		}
+		if controllerutil.RemoveFinalizer(fresh, acSecretFinalizer) {
+			if err := helperObj.GetClient().Update(ctx, fresh); err != nil {
+				return fmt.Errorf("failed to remove protection finalizer from %s: %w", s.Name, err)
+			}
+			logger.Info("Removed protection finalizer from AC secret", "secret", s.Name)
+		}
+		if err := helperObj.GetClient().Delete(ctx, fresh); err != nil && !k8s_errors.IsNotFound(err) {
```
fmount: Do we need the Delete? I'm wondering if `RemoveFinalizer` + `Update` is enough to see the `Secret` deleted.

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -346,44 +389,264 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+func (r *ApplicationCredentialReconciler) cleanupUnusedRotatedSecrets(
+	ctx context.Context,
+	instance *keystonev1.KeystoneApplicationCredential,
+	helperObj *helper.Helper,
+	identClient *gophercloud.ServiceClient,
+	userID string,
+) error {
+	logger := r.GetLogger(ctx)
+	serviceName := strings.TrimPrefix(instance.Name, "ac-")
+	protected := make(map[string]bool, 2)
```
fmount: do you need a `protected` map here?
I think (just theory) you can remove from L596 to L602 and simply have on L617 (where `protected` is used):

```golang
if s.Name == instance.Status.SecretName || s.Name == instance.Status.PreviousSecretName || hasConsumerFinalizer(s) {
...
```
it might be a cleaner approach as we do not inference (when we read the code) the hidden detail about string comparison.

internal/controller/keystoneapplicationcredential_controller.go
Deydra71: Yes, the description is correct. 

`openstack.org/ac-secret-protection` is set by keystone op at creation that prevents the secret from being deleted before the AC is revoked in Keystone. The current secret and previous used secret are protected by this finalizer.

`openstack.org/<service>-ac-consumer` is set by the service operator while it's actively using that secret.

internal/controller/keystoneapplicationcredential_controller.go
Deydra71: The secret still has owner reference to the AC CR that exists, so it wouldn't be collected unless the CR is deleted. For security reasons we don't want to keep unused AC secrets laying around, so we want to revoke AC in Keystone when these are fulfilled:

- it's not current `SecretName`
- it's not `PreviousSecretName`
- it doesn't have consumer finalizer

internal/controller/keystoneapplicationcredential_controller.go
Deydra71: I didn't use it because the lib-common `GetSecret` computes hash we don't need here. So I chose the raw get. But I see now that in other operators we generally use it even when the hash is just discarded, so I will do that too too keep the pattern.

internal/controller/keystoneapplicationcredential_controller.go
Deydra71: We actually already have it - https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/secret/secret.go#L97 , thanks!

internal/controller/keystoneapplicationcredential_controller.go
Deydra71: At first I thought that we should re Get to guard against stale `resourceVersion` on Update, because between List on #604 and Update on #636 we call `revokeKeystoneAC` and in the meantime of that external call some service operator could modify the secret changint the `resourceVersion`. 

However if we already simplify it to use `GetSecrets()` the List result is going to be fresh, so I don't think it's necessary with the other agreed changes.

Deydra71: Actually, I think that even if the conflict would happen the reconcile will retry again, because if  conflict happens then it means there's some change on the resource and Owns() watch is triggered and enqueue new recopncile

internal/controller/keystoneapplicationcredential_controller.go
Deydra71: Yes, totally agree with you. The map is an overkill

internal/controller/keystoneapplicationcredential_controller.go
Deydra71: There's sadly no available call to revoke multiple ACs in one API call, so the wrapper would jsut loop internally over the list of ACs meant to be revoked, but at the end revoking them one by one anyway. 

But with the other changes you sugegsted I think we simplified the reconcileDelete pretty well:
-   merged the revocation loop and the finalizer removal loop into one pass
- replaced `client.List` with `oko_secret.GetSecrets` (one call)
- building the Keystone client once upfront and not nesting the whole revocation logic inside the `keystoneAPI != nil` block

Please let me know if it makes more sense now, or if it's there some additional simplification/sugegstion.

fmount: Thanks, very good progress so far, I'll take another pass during the week!

Deydra71:
In latest chjanges addressed two found issues:
* `SetupWithManager` now checks RESTMapping before registering the OpenStackDataPlaneNodeSet watch
* reconcileDelete now retries on `AreSecretHashesInSync`  - the `!inSync` case is normally recovered by the NodeSet watch, but if the watch is absent (CRD not installed) or `AreSecretHashesInSync` hits a transient error the CR could be stuck in deletion

Deydra71:
/test functional

All passed locally

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -335,7 +369,23 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+		// Return early after creation/rotation so the defer patches the status.
+		// The status patch triggers a re-reconcile via the For() watch, at which
+		// point the cache is fresh and cleanup of unused rotated secrets can run.
+		// RequeueAfter is a safety net in case the watch event is delayed.
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
+	}
+
+	// Migrate old mutable secrets: add the application-credential-service label
+	// if missing, so they become visible to label-based cleanup and deletion queries
+	serviceName := strings.TrimPrefix(instance.Name, "ac-")
```
stuggi: this is used in 5 places. I think it would be good to make a func next to  `GetACCRName` to do the reverse, like `GetServiceNameFromAC` or so.

internal/controller/keystoneapplicationcredential_edpm_test.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	result, err := reconciler.reconcileDelete(context.Background(), instance, helperObj, nil)
+	if err != nil {
+		t.Fatalf("reconcileDelete returned error: %v", err)
+	}
+
+	// The controller relies on the NodeSet watch to trigger the next reconcile
+	// when EDPM deploys complete — no RequeueAfter needed
+	if result != (ctrl.Result{}) {
```
stuggi: but isn't it returning `ctrl.Result{RequeueAfter: time.Minute}` if we wait for a NodeSet ?

stuggi:
as discussed, the AC or AC secret consumed for edpm should get a label or annotation to reflect this. with this we could remove the global wait for edpm nodes to be update before cleanup can happen and ctlplane only services ac secret can be deleted as soon it is released.

Deydra71:
@stuggi addressed those two comments and added check for edpm-service annotation, so we don't block non-edpm services. Thanks!

api/v1beta1/keystoneapplicationcredential.go
```diff
@@ -18,70 +18,114 @@ package v1beta1
...
+	// EDPM nodes. The AC controller gates cleanup and deletion on NodeSet
+	// secret hash sync when this annotation is "true".
+	EDPMServiceAnnotation = "keystone.openstack.org/edpm-service" // #nosec G101
 )
 
-// GetApplicationCredentialFromSecret fetches and validates AC data from the Secret
-func GetApplicationCredentialFromSecret(
+// IsEDPMService returns true if this AC CR is annotated as an EDPM service.
+func (ac *KeystoneApplicationCredential) IsEDPMService() bool {
+	return ac.GetAnnotations()[EDPMServiceAnnotation] == "true"
```
stuggi: I am wondering for saftey mechanism, if we should return true if the annotation is set and true, or when the annotation is missing. so only return false if the annotation is there and false. what you think?

internal/controller/keystoneapplicationcredential_controller.go
```diff
@@ -346,44 +396,274 @@ func (r *ApplicationCredentialReconciler) reconcileNormal(
...
+
+	// Revoke ACs in Keystone for each secret that is not the current or previous secret and does not have a consumer finalizer
+	for i := range secretList.Items {
+		s := &secretList.Items[i]
+		if s.Name == instance.Status.SecretName || s.Name == instance.Status.PreviousSecretName || hasConsumerFinalizer(s) {
+			continue
+		}
+
+		acID := string(s.Data[keystonev1.ACIDSecretKey])
+		if acID != "" {
```
stuggi: should be ``if acID != "" && identClient != nil`` I think?

docs/applicationcredentials.md
```diff
@@ -200,43 +216,50 @@ This triggers seamless rotation with one pod restart and no authentication fallb
...
+
+**How it works:**
+- Before cleaning up unused rotated secrets **or proceeding with AC CR deletion**, the controller calls `edpm.AreSecretHashesInSync()` from `lib-common/modules/edpm`. This compares each `OpenStackDataPlaneNodeSet`'s `status.secretHashes` against the live secret hashes in the namespace.
+- If any NodeSet has stale hashes (config changed but EDPM not yet redeployed), **all** AC secret cleanup and AC CR deletion are deferred. This is a coarse-grained, fail-safe gate.
+- The controller watches `OpenStackDataPlaneNodeSet` resources (via unstructured access, no typed import) and re-evaluates eligibility when a NodeSet status changes — typically after an EDPM deploy updates `status.secretHashes`.
+- If the NodeSet CRD is not installed or no NodeSets exist, cleanup and deletion proceed normally.
+
+**Why all AC CRs are checked (not just EDPM services):**
+The controller checks all `KeystoneApplicationCredential` CRs uniformly against NodeSet hashes rather than maintaining a hardcoded mapping of which AC CRs correspond to EDPM services. This keeps keystone-operator generic and avoids coupling to specific service names.
```
stuggi: with the opt-in annotation this is no longer accurate? it should be either if the annotation is true, or with the above safety fallback if the annotation is true or not present?

api/v1beta1/keystoneapplicationcredential.go
Deydra71: As a good guard mechanism to not revoke EDPM related AC when the annotation is deleted or missing for some reason. So, we will revoke only when explicitly set to false.

That's good, I will update the https://github.com/openstack-k8s-operators/openstack-operator/pull/1922 because right now we were setting true only on edpm services and for rest the annotation was missing.

docs/applicationcredentials.md
Deydra71: Yes, I forgot to update the docs, thanks!

Deydra71:
Did a version mistake when rebasing, now the dependencies from lib-common to test with dataplane CRD should be correct.

mauricioharley (approved):
Let's go!

PR #693, opened 2026-04-21
Use OpenStack CR operatorOverrides to scale down keystone operator

karelyatin:
Switch from CSV-based scaling to using OpenStack CR operatorOverrides
  feature for scaling down the keystone operator. This avoids the need
  to disable the openstack init operator and provides better integration
  with the OpenStack operator management system.


abays (approved):
/lgtm

PR #694, opened 2026-04-22
Replace hardcoded namespace in KUTTL tests with $NAMESPACE variable

abays:
KEYSTONE_KUTTL_NAMESPACE is user-configurable in install_yamls (via ?= assignment), but test scripts and asserts had "keystone-kuttl-tests" hardcoded in service hostnames, secret metadata, and databaseHostname assertions. Tests would fail when run with a non-default namespace.

Replace all hardcoded namespace references with the $NAMESPACE env var that KUTTL provides to test scripts. For databaseHostname validation, switch from static YAML asserts to script-based checks that dynamically construct the expected value.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/test/kuttl/common/assert_sample_deployment.yaml
+++ b/test/kuttl/common/assert_sample_deployment.yaml
@@ -32,7 +32,6 @@ spec:
       memory: 500Mi
   secret: osp-secret
 status:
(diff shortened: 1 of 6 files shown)


Deydra71 (approved):
/lgtm

PR #706, opened 2026-05-18
Consume common templates via CommonTemplates field

fmount:
The `ssl.conf` `httpd` configuration is duplicated across service operators.
This patch removes the local copy and consumes it from [lib-common](https://github.com/openstack-k8s-operators/lib-common/pull/694).

Jira: [OSPRH-30239](https://redhat.atlassian.net/browse/OSPRH-30239)


abays (approved):
/lgtm

PR #707, opened 2026-05-25
Custom httpd configuration support

fmount:
Introduce a new sample demonstrating how to customize `httpd` conf using `extraMounts`. 
This approach leverages the `extraMounts` feature to inject custom configuration files loaded from `/etc/httpd/conf_custom`, enabling users to modify `httpd` settings without introducing new API parameters to the operator.
This follows the pattern already present in storage operators (e.g. https://github.com/openstack-k8s-operators/cinder-operator/pull/610)

### DIFF
--- a/config/samples/httpd-overrides/README.md
+++ b/config/samples/httpd-overrides/README.md
@@ -1,0 +1,121 @@
+# Keystone HTTPD Configuration Overrides
+
+The keystone-operator provides mechanisms to customize the Apache HTTPD server
+configuration through the use of custom configuration files. This feature
+leverages the
+[ExtraMounts](https://github.com/openstack-k8s-operators/dev-docs/blob/main/extra_mounts.md)
+functionality to mount custom HTTPD configuration files into the Keystone
+deployment.
+
+## How It Works
+
+1. **Custom Configuration Files**: Create HTTPD configuration files with your
+   custom settings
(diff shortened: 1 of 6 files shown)


config/samples/httpd-overrides/README.md
```diff
@@ -0,0 +1,119 @@
...
+   custom settings
+2. **ConfigMap**: Create ConfigMaps from files containing the overrides
+3. **OpenStackControlPlane Patch**: Patch the control plane to mount the
+   generated ConfigMap into Keystone containers. The HTTPD configuration
+   automatically includes files mounted to `/etc/httpd/conf_custom/*.conf`
+
+
+### Step 1: Create Custom HTTPD Configuration
+
+Create your custom HTTPD configuration file(s). The filename must start with
```
Deydra71: I wonder where are we enforcing the file name to be `httpd_custom` to be included in httpd configuration, if the httpd template seems to be including all files ending with `*.conf` ?
```
IncludeOptional conf_custom/*.conf
```

fmount: You're right, `httpd_custom_` might just be a best practice, but in general it includes whatever we pass there. Let me amend this part in my next PS.

fmount: Done

Deydra71 (approved):

PR #709, opened 2026-06-09
ci: add openstack-meta-content-provider-master job

xek:
## Summary

- Add `keystone-openstack-meta-content-provider-master` job that builds fresh keystone containers from OpenDev master on each PR
- Add `keystone-operator-tempest-master` job that consumes those fresh containers
- Both jobs added to the `github-check` pipeline (tempest-master non-voting)

This allows co-submitting a keystone service fix and an operator change in a single review round, without waiting for the service fix to be packaged into `current-tested` first. Follows the pattern established by [telemetry-operator #894](https://github.com/openstack-k8s-operators/telemetry-operator/pull/894) and watcher-operator.

Jira: https://redhat.atlassian.net/browse/OSPRH-30892

Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -26,3 +26,42 @@
       cifmw_test_operator_tempest_concurrency: 3
       cifmw_test_operator_tempest_include_list: |
         ^tempest.api.identity.
+
+- job:
+    name: keystone-openstack-meta-content-provider-master
+    parent: openstack-meta-content-provider-master
+    description: |
+      A meta content provider zuul job to build keystone
+      specific containers from OpenDev master.
+    vars:
+      cifmw_build_containers_force: false
+      cifmw_bop_dlrn_from_source: true
+      cifmw_build_containers_registry_namespace: podified-master-centos10
+      cifmw_operator_build_operators:
+        - name: keystone-operator
+          src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/keystone-operator"
+        - name: openstack-operator
+          src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/openstack-operator"
+          image_base: keystone
+      cifmw_build_containers_image_tag: keystone_latest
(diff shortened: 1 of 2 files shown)


zuul.d/jobs.yaml
```diff
@@ -26,3 +26,42 @@
...
         ^tempest.api.identity.
+
+- job:
+    name: keystone-openstack-meta-content-provider-master
+    parent: openstack-meta-content-provider-master
+    description: |
+      A meta content provider zuul job to build keystone
+      specific containers from OpenDev master.
+    vars:
+      cifmw_build_containers_force: true
```
Deydra71: I think, given how keystone has frequent operator changes, we shouldn't trigger the run on every PR, increasing the run time and potential unrelated voting failures if something upstream is wrong.  Telemetry previously (before commit [ee3bb96](https://github.com/openstack-k8s-operators/telemetry-operator/commit/ee3bb96)) used `cifmw_build_containers_force: false` , IIUC the build only happened when a OpenDev Depends-On was present, what do you think @xek ?

mauricioharley: Agreeing with @Deydra71, and to be more specific: `keystone-operator-tempest-master` is already non-voting, but `keystone-openstack-meta-content-provider-master` is not. We should add `voting: false` to it as well in `zuul.d/projects.yaml`:

```
- keystone-openstack-meta-content-provider-master:
    voting: false
```

The co-submission workflow still works:  when a `Depends-On` to OpenDev is present, the job result is visible in the PR even as non-voting.

We can see this happen in practice in telemetry-operator's PR #894.  That PR required multiple rechecks over weeks precisely because of upstream issues in the `meta-content-provider` job.

mauricioharley (dismissed):

Deydra71 (commented):
let's squash the commits and then we are good to go

Deydra71 (approved):

mauricioharley (approved):

PR #712, opened 2026-06-16
Add transport secret consumer finalizer for credential rotation

lmiccini:
Add consumer finalizer management for transport URL secrets, modeled
after the existing ApplicationCredential consumer finalizer pattern.
This ensures the TransportURL controller waits for all consumers to
roll out with new credentials before releasing old RabbitMQ users
during rotation.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Depends-on: https://github.com/openstack-k8s-operators/infra-operator/pull/606


lmiccini:
recheck

PR #715, opened 2026-06-22
Fix shell injection in trust_flush CronJob

stuggi:
Replace the bash -c string concatenation of TrustFlushArgs with a direct argv array, preventing shell metacharacter injection. The CronJob now executes keystone-manage as Command: ["keystone-manage", "trust_flush", ...args] instead of Command: ["/bin/bash"] Args: ["-c", "keystone-manage trust_flush" + args].

Additionally add kubebuilder:validation:Pattern on the TrustFlushArgs CRD field to reject shell metacharacters at the API level.

Jira: [OSPRH-31632](https://redhat.atlassian.net/browse/OSPRH-31632)


Deydra71 (approved):
lgtm

PR #716, opened 2026-06-23
Add transport URL secret rotation with consumer finalizer

lmiccini:
## What this does

When infra-operator rotates a RabbitMQ transport URL (creating a new secret and user), consumer operators must hold a **consumer finalizer** on the old secret until all their pods have rolled out with the new credentials. Without this, infra-operator cleans up the old RabbitMQ user while pods are still connected with the old credentials, causing message-bus outages.

## Approach (updated)

> **Note:** this PR was reworked from the original grace-period design to an event-driven "applied input hash" design. The old approach released the finalizer after a fixed grace period based on the parent `Ready` condition, which was racy against the informer cache and could revoke credentials prematurely. The new approach releases the finalizer only once every child has provably rolled out.

1. Add a consumer finalizer to the current transport URL secret early in reconcile. `Status.TransportURLSecret` is set for first-time setup only (empty or unchanged); during rotation the status is updated solely by `FinalizeSecretRotation` at the end of reconcile.

2. Pass `transportURL.Status.SecretName` directly to sub-CR creation and config generation as a parameter — never read `Status.TransportURLSecret` for sub-CR specs.

3. Each child controller records an `AppliedInputSecretHash` in its status, set only after `statefulset.IsReadyForInput` / `deployment.IsReadyForInput` confirms — via an uncached API read — that the workload is fully rolled out with the expected `CONFIG_HASH`.

4. The parent mirrors a child's Ready condition only when its `Generation == ObservedGeneration` and `AppliedInputSecretHash` matches the current input hash; otherwise it sets the condition to Unknown.

5. Guard: `FinalizeSecretRotation` removes the consumer finalizer from the old secret only when every child reports the expected hash and is ready. The guard is computed from `Conditions.AllSubConditionIsTrue()` (not `IsReady()`), because `Conditions.Init()` resets the `Ready` condition to Unknown on every reconcile.

The same pattern applies to notification transport URL secrets and application-credential secrets where applicable.

## Dependency

Depends on the lib-common `IsReadyForInput` / `FinalizeSecretRotation` helpers (currently pinned via a `replace` to the fork commit while the lib-common PR is in review).

### DIFF
--- a/api/v1beta1/keystoneapplicationcredential.go
+++ b/api/v1beta1/keystoneapplicationcredential.go
@@ -67,6 +67,8 @@ func (ac *KeystoneApplicationCredential) IsEDPMService() bool {
 	return ac.GetAnnotations()[EDPMServiceAnnotation] != "false"
 }
 
+// Deprecated: Use object.ManageSecretConsumerFinalizer and
+// object.FinalizeSecretRotation from lib-common instead.
 // ManageACSecretFinalizer ensures consumerFinalizer is present on the AC secret
 // identified by newSecretName and absent from the one identified by
 // oldSecretName. It is a no-op when both names are equal.
@@ -108,6 +110,7 @@ func ManageACSecretFinalizer(
 	return object.ManageConsumerFinalizer(ctx, h, newObj, oldObj, consumerFinalizer)
 }
 
+// Deprecated: Use object.RemoveSecretConsumerFinalizer from lib-common instead.
 // RemoveACSecretConsumerFinalizer removes consumerFinalizer from the AC secret
 // identified by secretName. It is a no-op when secretName is empty or the
 // secret no longer exists.
--- a/internal/controller/keystoneapi_controller.go
+++ b/internal/controller/keystoneapi_controller.go
@@ -42,6 +42,7 @@ import (
 	job "github.com/openstack-k8s-operators/lib-common/modules/common/job"
 	labels "github.com/openstack-k8s-operators/lib-common/modules/common/labels"
 	nad "github.com/openstack-k8s-operators/lib-common/modules/common/networkattachment"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/object"
 	common_rbac "github.com/openstack-k8s-operators/lib-common/modules/common/rbac"
 	oko_secret "github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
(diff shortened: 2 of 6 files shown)


api/v1beta1/keystoneapplicationcredential.go
```diff
@@ -66,69 +57,3 @@ const (
 func (ac *KeystoneApplicationCredential) IsEDPMService() bool {
 	return ac.GetAnnotations()[EDPMServiceAnnotation] != "false"
 }
-
-// ManageACSecretFinalizer ensures consumerFinalizer is present on the AC secret
-// identified by newSecretName and absent from the one identified by
-// oldSecretName. It is a no-op when both names are equal.
-func ManageACSecretFinalizer(
```
stuggi, line 73: maybe we should keep those funcs until all other operators got updated and no longer consume them? otherwise we break the regular bumps if their changes not landed before.

api/v1beta1/keystoneapplicationcredential.go
```diff
@@ -66,69 +57,3 @@ const (
...
-		}
-	}
-
-	return object.ManageConsumerFinalizer(ctx, h, newObj, oldObj, consumerFinalizer)
-}
-
-// RemoveACSecretConsumerFinalizer removes consumerFinalizer from the AC secret
-// identified by secretName. It is a no-op when secretName is empty or the
-// secret no longer exists.
-func RemoveACSecretConsumerFinalizer(
```
stuggi, line 114: as above, maybe we mark them as deprecated in this PR

internal/controller/keystoneapi_controller.go
```diff
@@ -522,6 +523,11 @@ func (r *KeystoneAPIReconciler) reconcileDelete(ctx context.Context, instance *k
 		}
 	}
 
+	if err := object.RemoveSecretConsumerFinalizer(ctx, helper, instance.Namespace,
```
stuggi, line 527: only removes the consumer finalizer from instance.Status.TransportURLSecret, which during a pending TransportURL rotation still points at the old secret while the finalizer was already added to the new one. If the CR is deleted mid-rotation, the new secret's finalizer is never cleaned up, permanently stuck in Terminating if later deleted. Wondering if we should re-fetch the owned TransportURL CR and remove the consumer finalizer from its Status.SecretName too, alongside the existing instance.Status.TransportURLSecret cleanup?

api/v1beta1/keystoneapplicationcredential.go
lmiccini: good point, will keep them for now and send another pr later.

lmiccini:
recheck

stuggi (approved):
/lgtm

lmiccini:
recheck

stuggi (approved):
/lgtm

PR #723, opened 2026-06-29
Fix client IP logging in Keystone API VirtualHost

fmount:
The `VirtualHost` `CustomLog` directive was overriding the server-level dual-log setup, causing all requests to be logged with the OpenShift router IP instead of the real client IP from `X-Forwarded-For`.

Add `SetEnvIf` and dual `CustomLog` lines inside the `VirtualHost` block to match the `nova-operator` pattern.

Jira: [OSPRH-32109](https://redhat.atlassian.net/browse/OSPRH-32109)


stuggi (approved):
/lgtm

PR #728, opened 2026-07-14
[18.0-fr6] Fix shell injection in trust_flush CronJob

stuggi:
Replace the bash -c string concatenation of TrustFlushArgs with a direct argv array, preventing shell metacharacter injection. The CronJob now executes keystone-manage as Command: ["keystone-manage", "trust_flush", ...args] instead of Command: ["/bin/bash"] Args: ["-c", "keystone-manage trust_flush" + args].

Jira: [OSPRH-32064](https://redhat.atlassian.net/browse/OSPRH-32064)

### DIFF
--- a/internal/keystone/cronjob.go
+++ b/internal/keystone/cronjob.go
@@ -16,6 +16,8 @@ limitations under the License.
 package keystone
 
 import (
+	"strings"
+
 	memcachedv1 "github.com/openstack-k8s-operators/infra-operator/apis/memcached/v1beta1"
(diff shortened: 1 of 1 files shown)


Deydra71:
@stuggi a question - I assume we are not adding the kubebuilder pattern validation like in the original PR on purpose because of CR compatibility?

stuggi:
> @stuggi a question - I assume we are not adding the kubebuilder pattern validation like in the original PR on purpose because of CR compatibility?

yes, the  CRD changes are not a must for the fix and to not introduce it in the maintenance release I removed them

Deydra71 (approved):
/lgtm

PR #740, opened 2026-07-27
Restore pods RBAC needed for NetworkAttachments verification

stuggi:
Commit 2adbb94 ("Remove unnecessary pods RBAC permissions") dropped the pods kubebuilder RBAC marker from keystoneapi_controller.go, which calls lib-common's VerifyNetworkStatusFromAnnotation to verify NetworkAttachments. That function lists Pods using the controller-manager's own client, so the manager's ClusterRole (config/rbac/role.yaml) needs get;list on pods or KeystoneAPI gets stuck in NetworkAttachmentsReady=False with a "pods is forbidden" error, as seen on manila/cinder/neutron-operator after the same cleanup. Restore the get;list marker and regenerate config/rbac/role.yaml. The unused rbacRules grant on the workload service account remains removed, since that really is unused.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -28,6 +28,13 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - ""
+  resources:
   - secrets/finalizers
   verbs:
   - create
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

stuggi:
recheck

PR #745, opened 2026-08-03
Fernet RNG should not return empty string

Deydra71:
`GenerateFernetKey` logged RNG failure then returned "". Callers stored that into the Fernet/credential Secret. Change the API to (string, error). On `rand.Read` failure, return a wrapped `error. ensureFernetKeys ` propagates it so reconcile fails and retries instead of writing bad secrets.


stuggi (approved):
/lgtm

PR #747, opened 2026-08-06
Remove kolla entrypoint pattern, harden SecurityContext

stuggi:
Eliminate kolla_start/kolla_set_configs and the config.json staging pattern. Configs are now mounted directly to final paths via SubPath volume mounts, removing the need for root privilege escalation at container startup.

Key changes:
- Replace kolla_start with /usr/sbin/httpd -DFOREGROUND
- Remove kolla_set_configs from bootstrap command
- Remove all KOLLA_CONFIG_STRATEGY and KOLLA_BOOTSTRAP env vars
- Delete keystone-api-config.json (no longer needed)
- Mount configs via SubPath to /etc/keystone/, /etc/httpd/, /etc/my.cnf
- Add emptyDir volumes for writable paths (/run/httpd, /tmp, /var/log) using lib-common volume.WritableDirVolume/Mount helpers with standardized constants (volume.RunHttpdVolumeName, TmpVolumeName, etc.)
- Use pod.RestrictiveSecurityContext and RestrictivePodSecurityContext from lib-common for all workloads (deployment, bootstrap, dbsync, cronjob), setting ReadOnlyRootFilesystem, dropping ALL capabilities, and enabling seccomp RuntimeDefault
- Set AutomountServiceAccountToken=false on all workloads — none need k8s API access (ServiceAccountName is kept for SCC binding)
- Import KeystoneUID from lib-common modules/serviceuser
- Override TLS CertMount/KeyMount to /etc/pki/tls/ for Apache
- Migrate SCC from anyuid to nonroot-v2
- Update httpd.conf: User/Group keystone, PidFile /run/httpd/httpd.pid
- Support custom httpd config keys via dynamic SubPath mounts
- Config Secret DefaultMode set to 0440 (owner-read + group-read, most restrictive — all SubPath mounts are ReadOnly: true anyway)
- Update functional and kuttl tests

Testing on a cluster surfaced a permission denial starting httpd:

  httpd: Syntax error on line 16 of /etc/httpd/conf/httpd.conf:
  Could not open configuration file /etc/httpd/conf.d/auth_openidc.conf:
  Permission denied

mod_auth_openidc's auth_openidc.conf ships root:apache 0640 and is baked into the container image rather than volume-mounted, so FSGroup does not apply to it. Under the previous kolla/anyuid setup this was never an issue: the httpd master process ran as root and could read any file regardless of group. TCIB never adds the keystone service user to the apache group (uid_gid_manage.sh only ever adds it to "kolla"), so running the whole pod as non-root keystone requires the apache group to be granted explicitly at the pod level via RestrictivePodSecurityContext's new supplementalGroups parameter and lib-common's serviceuser.ApacheGID constant.

Testing further surfaced httpd failing to start with:

  (13)Permission denied: AH00091: httpd: could not open error log
  file /etc/httpd/logs/error_log.
  AH00015: Unable to open logs

/etc/httpd/logs is a symlink to /var/log/httpd (root:root 0700 in the image). httpd.conf only redirected ErrorLog/CustomLog to /dev/stdout inside <VirtualHost> blocks, so the global/startup error log fell back to Apache's compiled-in default of "logs/error_log" relative to ServerRoot. Fixed by adding a top-level ErrorLog directive, plus a var-log-httpd emptyDir mounted at /var/log/httpd as defense-in-depth for any other RPM-shipped conf.d file that references relative "logs/*" paths.

ErrorLog (both the top-level and per-vhost directives) is sent to /dev/stderr rather than /dev/stdout, following the convention used by the official Apache httpd container image and OpenShift Logging's per-stream labeling, keeping error diagnostics distinguishable from routine access logs (CustomLog stays on /dev/stdout).

Testing also surfaced a keystone WARNING:

  keystone.common.fernet_utils [...] key_repository is world readable:
  /etc/keystone/fernet-keys

The fernet-keys and credential-keys Secret volumes had no DefaultMode set, so Kubernetes defaulted to 0644 (world-readable), unlike config-data which was already hardened to 0440. Fixed by setting DefaultMode 0440 on both, matching config-data and silencing the warning.

Jira: [OSPRH-33504](https://redhat.atlassian.net/browse/OSPRH-33504)
Jira: [OSPRH-33503](https://redhat.atlassian.net/browse/OSPRH-33503)
    
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/728


stuggi:
recheck

internal/controller/keystoneapi_controller.go
```diff
@@ -1345,8 +1345,27 @@ func (r *KeystoneAPIReconciler) reconcileNormal(
...
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get httpd custom config secret %s: %w",
+				*instance.Spec.HttpdCustomization.CustomConfigSecret, err)
+		}
+		if len(overrideSecret.Data) > 0 {
+			for _, endpt := range []service.Endpoint{service.EndpointInternal, service.EndpointPublic} {
+				for _, key := range slices.Sorted(maps.Keys(overrideSecret.Data)) {
+					if len(overrideSecret.Data[key]) > 0 {
+						customConfigKeys = append(customConfigKeys, "httpd_custom_"+endpt.String()+"_"+key)
+					}
```
abays: Is it possible that there could be a situation where we haven't produced both endpoint-prefixed variants for every key?  I think pod startup would fail in such a scenario.

stuggi: yes, good catch. there might be an issues, since we fetch it twice and there might be a double-fetch race. generateServiceConfigMaps now returns customConfigKeys derived directly from the same customTemplates map it renders into the config Secret, instead of reconcileNormal re-fetching HttpdCustomization.CustomConfigSecret independently and recomputing the key list. With this we get a single source of truth and no more divergence if there is one. I think this should address the issue?

internal/keystone/bootstrap.go
```diff
@@ -66,7 +65,7 @@ func BootstrapJob(
 	// create Volume and VolumeMounts
 	bootstrapExtraMounts := []keystonev1.KeystoneExtraMounts{}
 	volumes := getVolumes(instance, bootstrapExtraMounts, BootstrapPropagation)
-	volumeMounts := getVolumeMounts(bootstrapExtraMounts, BootstrapPropagation)
+	volumeMounts := getBootstrapVolumeMounts()
```
abays, line 68: This never mounts the `scripts` volume that gets returned by L67.  Is that intentional for now?

internal/controller/keystoneapi_controller.go
abays: Sounds good!

internal/keystone/deployment.go
```diff
@@ -80,13 +82,18 @@ func Deployment(
...
-	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
 	envVars["CONFIG_HASH"] = env.SetValue(configHash)
 
 	// create Volume and VolumeMounts
 	volumes := getVolumes(instance, instance.Spec.ExtraMounts, KeystonePropagation)
 	volumeMounts := getVolumeMounts(instance.Spec.ExtraMounts, KeystonePropagation)
 
+	volumes = append(volumes,
+		volume.WritableDirVolume(volume.RunHttpdVolumeName),
+		volume.WritableDirVolume("var-log-keystone"),
```
abays: Maybe we could use a constant for `var-log-keystone`, since it is used both here and in `volumes.go`?

internal/keystone/bootstrap.go
stuggi: there are no scripts . templates/keystoneapi/bin/init.sh was the only file ever present, and it was deleted in commit fe0776d ("Render service config in operator and store in secret" / "Removes the need of an
  init container, renders all configs in go and stores the resulting configs on a secret"). I'll remove scripts in all places.

internal/keystone/deployment.go
stuggi: sure

stuggi: done

internal/keystone/bootstrap.go
stuggi: removed

stuggi:
recheck

abays (approved):
/lgtm

Deydra71 (approved):
/lgtm

PR #767, opened 2026-09-01
[CI] Speculative S2I keystone deploy via OpenStackVersion

Deydra71:
Jobs added to github-check (non-voting siblings; existing kuttl/tempest unchanged):

- keystone-s2i-content-provider — `child of s2i-openstack-container-content-provider`; builds `keystone/keystone` (maps to `keystoneAPIImage` in OpenStackVersion)

- keystone-s2i-test — child of `s2i-test-base`; deploys s2i-built images via set_containers during edpm_prepare and runs test-operator with:
  - include: `^tempest.api.identity`.
  - `tempestconf` overrides from the downstream base component job (`identity-feature-enabled.application_credentials`, `identity-feature-enabled.project_tags`, `identity-feature-enabled.enforce_scope`)
  - 
Pipeline: keystone-s2i-test depends on both openstack-k8s-operators-content-provider (operator catalog) and keystone-s2i-content-provider (s2i service images).

Depends-On: openstack-k8s-operators/s2i-openstack-containers#178

Reference: https://github.com/openstack-k8s-operators/s2i-openstack-containers/blob/main/docs/operator-onboarding.md

PR #776, opened 2026-09-15
[DNM] Pin keystone-s2i-test to S2I-built openstack-tempest

Deydra71:
Build tempest in keystone-s2i-content-provider and override `cifmw_test_operator_tempest_*` from s2i_ci_content.


Deydra71:
This PR is to force tempest from main and see if it resolves tempest failures due to using old (Antelope) version as in https://github.com/openstack-k8s-operators/keystone-operator/pull/767

